PR #149, opened 2022-04-17
Added screenshot gif

adrianwitaszak:


bitspittle (changes requested):
Hey thanks for the nice animation! And sorry for only catching this just now.

By the way, the official templates are here: https://github.com/varabyte/kobweb-templates/tree/main/examples/clock in case you were expecting this animation to show up after this commit.

Not submitting this as is because I think we can instead host this asset in the varabyte media repo. I've been doing that with Kotter which is nice because the code is still really really small to sync the project while all the samples refer to the media repo indirectly which itself is many many multiple MB at this point.

For example check out my example landing page readme here: https://github.com/varabyte/kotter/tree/main/examples/doomfire

What do you think about first creating a pull request against https://github.com/varabyte/media and putting this clock asset in, say, kobweb/images/templates/clock/ and then updating this pull request to reference that url?

adrianwitaszak:
created PR in media:
https://github.com/varabyte/media/pull/1
this one can be closed

PR #151, opened 2022-05-21
Support custom name in source sets

### LINKED ISSUE #150: Cannot build project with custom source sets name
In Kotlin Multiplatform is it possible to change the default source set name (like `js` or `jvm`) to a custom one

`build.gradle.kts`
```
kotlin {
    jvm("myBackend") {...}
    js("myFrontend", IR) {...}
    sourceSets {
        val myFrontendMain by getting {...}
        val myBackendMain by getting {...}
    }
}
```

When running `kobwebStart` the following error occurs:
`Task with name 'jsBrowserDevelopmentRun' not found in root project`

DVDAndroid:
In Kotlin Multiplatform is it possible to change the default source set
name (like `js` or `jvm`) to a custom one

This commit replaces all hardcoded task names and source set references
with its corresponding exact values

Fixes #150


DVDAndroid:
Used examples as build test

bitspittle (approved):
Wow, that's a great refactoring. Thank you for taking the time!!

Issue #155, opened 2022-06-21
Add support for Material Design icons


bitspittle:
Like Font Awesome icons, we could publish a library with Material Design icons.

See also: 

* https://fonts.google.com/icons?selected=Material+Icons+Round
* https://materialdesignicons.com/

Marking this as good first issue since it's (hopefully) somewhat straightforward using the [kobweb-silk-icons-fa](https://github.com/varabyte/kobweb/tree/main/frontend/kobweb-silk-icons-fa) module as reference.

PR #191, opened 2022-12-03
Add support for Material Design icons 

jassycliq:
Implements Material Design Icon support as requested in #155.
Still playing around with this work but it's working as is.


frontend/kobweb-silk-icons-mdi/README.md
```diff
@@ -0,0 +1,27 @@
+Support for integration of font awesome icons in your project.
```
bitspittle, line 1: Just a heads up that this README will need to be updated (or you can delete it for now and I can add it in later).

One thing that would be really helpful for me is if you shared your instructions for how you created `mdi-icon-list.txt`, so that in the future I can replicate your work if users ask for an updated list of Material Design icons.

frontend/kobweb-silk-icons-mdi/build.gradle.kts
```diff
@@ -0,0 +1,182 @@
...
+        }
+    }
+}
+
+kobwebPublication {
+    artifactId.set("kobweb-silk-icons-mdi")
+    description.set("A collection of Kobweb Silk components that directly wrap Material Design icons")
+}
+
+enum class IconCategory {
```
bitspittle, line 32: I'm looking at https://fonts.google.com/icons which seems to have two categories - material symbols and material icons. If we're using that approach, it looks like there should be two enumerations - SymbolStyle (Outlined, Round, Sharp) and IconStyle (Outlined, Filled, Round, Sharp, Two Tone)

It seems like some names span both categories (e.g. there's a search icon *and* a search symbol), so it seems like we'd probably want to use different prefixes for each type.

I'm wondering if we'd want an API like this:

```
MdsSearch(modifier: Modifier = Modifier, style: SymbolStyle? = null)
MdiSearch(modifier: Modifier = Modifier, style: IconStyle? = null)
```

**This is assuming you're looking at the same source of the icons as I am**. I can appreciate you might be using a different place to get all icons, e.g. https://materialdesignicons.com/ which is not (as far as I can tell) the official Google version but it is also the first search result for "Material Design Icons" so :shrug:

frontend/kobweb-silk-icons-mdi/build.gradle.kts
```diff
@@ -0,0 +1,182 @@
...
+
+package com.varabyte.kobweb.silk.components.icons.mdi
+
+import androidx.compose.runtime.*
+import com.varabyte.kobweb.compose.ui.Modifier
+import com.varabyte.kobweb.compose.ui.toAttrs
+import org.jetbrains.compose.web.dom.Span
+import org.jetbrains.compose.web.dom.Text
+
+enum class IconCategory(internal val className: String) {
```
bitspittle, line 119: It doesn't look like we need an IconCategory in material design. That exists because that's how font awesome asks you to create their icons.

In font awesome, the idea was that 99%+ of users would use `IconStyle` instead and never `IconCategory`. `IconStyle` is a simpler, more restrictive API.

However, `IconCategory` is still provided just in case I happened to be missing some icon for some reason (maybe a new release of font awesome icons get dropped that I never update Kobweb for)

But here, your `IconCategory` and `IconStyle` look like they're identical. Better then to just have one enum.

Here's what I'm currently thinking (using an extension method since we also need to handle the null case):

```
enum class IconStyle {
   FILLED,
   OUTLINED,
   ROUND,
   SHARP,
   TWO_TONE,
}

fun IconStyle?.toClassName(): String {
   return buildString {
      append("material-icons")
   when (this) {
      OUTLINED -> append("-outlined")
      ROUND -> append("-round")
     ....
   }
}
```

And based on my earlier review comment, we'd also need to do something similar for `SymbolStyle` (assuming that previous comment was even correct)

Let me know if this makes sense!

bitspittle (commented):
Just wanted to say again, this is amazing. Thanks for kickstarting it!

There's a few comments below, but the overall theme is I'd love to know where you're getting your icons from. I'm assuming https://fonts.google.com/icons but you might be using https://materialdesignicons.com/ ?

In general, when doing code reviews with me, please never feel like I'm a dictator telling you what to do blindly. You can always reply with pushback or questions.

Finally, feel free to put "Fixes #155" at the bottom of your commit message, since with what you have here I think we could definitely consider that issue closed.

bitspittle:
By the way, you can also introduce a "Default" style as an alternate to a nullable style. I'm fine with both approaches in case you like one over the other.

jassycliq:
> Just wanted to say again, this is amazing. Thanks for kickstarting it!
> 
> There's a few comments below, but the overall theme is I'd love to know where you're getting your icons from. I'm assuming https://fonts.google.com/icons but you might be using https://materialdesignicons.com/ ?
> 
The original work that I had shown you was the latter website but this PR is based off https://fonts.google.com/icons.

> In general, when doing code reviews with me, please never feel like I'm a dictator telling you what to do blindly. You can always reply with pushback or questions.
>
👍🏽 

> Finally, feel free to put "Fixes #155" at the bottom of your commit message, since with what you have here I think we could definitely consider that issue closed.
>
👍🏽

frontend/kobweb-silk-icons-mdi/build.gradle.kts
jassycliq: Thanks for those hints. I was originally looking at having the Symbols and Icons be separate work but makes sense to keep them together 👍🏽 

I've made some changes to progress towards the new API but still got plenty of work to do (a.k.a Gradle script)

frontend/kobweb-silk-icons-mdi/build.gradle.kts
jassycliq: I've made some changes here based on your comments. 
I realized that `SymbolStyle`s all share the same icons whereas `IconStyle`s may contain different icons so I believe the changes I've made will handle these differences.

Please let me know if you can see any issues 😄

frontend/kobweb-silk-icons-mdi/build.gradle.kts
```diff
@@ -111,29 +111,28 @@ package com.varabyte.kobweb.silk.components.icons.mdi
...
+import com.varabyte.kobweb.silk.components.icons.mdi.MdStyle.SymbolStyle.Outlined
 import org.jetbrains.compose.web.dom.Span
 import org.jetbrains.compose.web.dom.Text
 
-sealed class MdCategory {
-    sealed class Symbol : MdCategory() {
-        object Outlined : Symbol()
-        object Rounded : Symbol()
-        object Sharp : Symbol()
+sealed class MdStyle {
```
bitspittle, line 123: Does this need to be a sealed class, instead of just two separate classes?

jassycliq: Good point. I was very much in the mindset of how it was being used in the gradle file.

frontend/kobweb-silk-icons-mdi/build.gradle.kts
```diff
@@ -111,29 +111,28 @@ package com.varabyte.kobweb.silk.components.icons.mdi
...
     }
 
-    sealed class Icon : MdCategory() {
-        object Default : Icon()
-        object Outlined : Icon()
-        object Rounded : Icon()
-        object Sharp : Icon()
-        object TwoToned : Icon()
+    sealed class IconStyle(internal val className: String?) : MdStyle() {
+        object Default : IconStyle(null)
```
bitspittle, line 131: Let's use a classic enum class here. See also: https://kt.academy/article/ek-enum 

TL;DR - Enum classes are simpler. If all you're doing is listing an simple enumeration of values, they should still be your go-to solution.

Enum classes are *slightly* more memory efficient than sealed classes, because you only create one instance of a class. So, in our bytecode, we'd have a single class (IconStyle) vs. 6 (IconStyle, IconStyle.Default, IconStyle.Outline...).

Also, the bytecode may be able to take advantage of the fact it realizes those values are all constants? I'm not sure.

We could always migrate an enum class to a sealed class later if we *had* to.

frontend/kobweb-silk-icons-mdi/build.gradle.kts
```diff
@@ -111,29 +111,28 @@ package com.varabyte.kobweb.silk.components.icons.mdi
...
+        object Sharp : SymbolStyle("sharp")
     }
 
-    sealed class Icon : MdCategory() {
-        object Default : Icon()
-        object Outlined : Icon()
-        object Rounded : Icon()
-        object Sharp : Icon()
-        object TwoToned : Icon()
+    sealed class IconStyle(internal val className: String?) : MdStyle() {
```
bitspittle, line 130: I would suggest keeping the `className` out of this core class and using an extension method instead.

Here's the blog post that inspired me (by Roman Elizarov, the Kotlin project lead): https://elizarov.medium.com/extension-oriented-design-13f4f27deaee

Here's what it looks like in this case:

```
enum class IconStyle {
   DEFAULT,
   OUTLINED,
   ...
}

private fun IconStyle.toClassNameSuffix(): String {
   return when(this) {
      DEFAULT -> ""
      OUTLINED -> "-outlined"
      ...
   }
}
```

Even though technically here the end code is the same, this emphasizes our intention as developers to what part of the classes we think of as the minimal representation, vs. say someone who just keeps adding more and more stuff into a class because it's a convenient place to put it!

If you start to apply this philosophy more and more to your code, you'll find that you might actually have the same simple class in one place extended *differently* in different modules, which is kind of neat.

bitspittle, line 130: Nit: Let's call this "classNameSufix" instead of "className"

bitspittle (changes requested):
The comments are long but not because of you, because of me. I don't know how to talk concisely. The code changes the comments recommend are very minimal.

frontend/kobweb-silk-icons-mdi/build.gradle.kts
jassycliq: 🤯 Gives me composition vs inheritance vibes

frontend/kobweb-silk-icons-mdi/build.gradle.kts
```diff
@@ -111,29 +111,28 @@ package com.varabyte.kobweb.silk.components.icons.mdi
...
+        object Default : IconStyle(null)
+        object Outlined : IconStyle("-outlined")
+        object Rounded : IconStyle("-rounded")
+        object Sharp : IconStyle("-sharp")
+        object TwoToned : IconStyle("two-toned")
```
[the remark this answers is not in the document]
jassycliq: They're working.
They don't exist for symbols though but that'll be a non-issue 😄

frontend/kobweb-silk-icons-mdi/build.gradle.kts
bitspittle: It definitely embraces the spirit of delegation. 👍

Warning: some programmers, especially older ones, hate hate hate extension methods, so you might find pushback at work around this concept. I've found Java devs in particular find the whole feature foreign and it scares them.

frontend/kobweb-silk-icons-mdi/build.gradle.kts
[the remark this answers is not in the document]
bitspittle: Another reason to boot symbols, hooray! Sorry again for wasting your time making you explore them.

jassycliq:
Almost didn't make my self-imposed timeline 😅

bitspittle:
> Almost didn't make my self-imposed timeline sweat_smile

Really appreciate it.

Timing is good, I'm planning on putting out a new Kobweb release this weekend (for the 1.2.2 change). I'll be sure to get this change in.

If you don't mind, I'm thinking about taking your pull request and smashing it into a single commit (keeping you as the author of course)

Got some morning chores, but will review later today!

PR #243, opened 2023-05-21
Added KMP example

charlee-dev:
I added a KMP example with Android, iOS, Desktop and Web (using Kobweb).


charlee-dev:
@bitspittle Or if you find a place, you can just link it to this repo. Then I can keep it updated and expand it myself.
https://github.com/charlee-dev/SAMPLE-KMP_SharedUI

bitspittle:
Hey Adrian! I replied in the Discord but I'll add a note here as well for completion.

First of all, if you spent any time on this PR at all, please accept my apologies. I was originally thinking you were going to add some sort of Kobweb example to your own and were asking me if I wanted to link to it.

When I originally started Kobweb, I *did* intend for this to become a growing collection of examples (like I do in Kotter: https://github.com/varabyte/kotter/tree/main/examples).

However, I found over time with Kobweb that it was a different beast and maintaining multiple examples was a PITA. More and more, I moved examples over into the kobweb-templates directory instead, which feels like a better home for them.

However, I always wanted to keep around one project (helloworld) because I used it to quickly experiment / test APIs with (since it uses a Gradle Composite build setup, which means if I run `kobweb run` inside of it, it automatically finds and builds all dependencies as necessary -- very convenient for any Kobweb dev!)

I kind of forgot about this, but after your PR, I realized that I should be more explicit with what I'm doing. So if you check the latest branch (0.13.5 at the time of writing this response), you'll see I killed the "examples" directory and created a "playground" directory instead, where I moved the old "examples/helloworld" project into it.

Hope this is more information than you cared to know :D However, I will open a bug that at *some* point there should be a KMP demo project which has at least Android *and* Kobweb running side by side.

PR #261, opened 2023-06-29
Fix incorrect modifier return value

DennisTsar:


bitspittle (approved):
Wow, subtle :)

Definitely one of the cases where me leaving off explicit types gets me into trouble. This won't be the last time.... Thanks for catching this one.

PR #263, opened 2023-07-02
Rework grid line names to be standalone values when configuring rows/cols

DennisTsar:
There are now three distinct subtypes of `GridEntry`: `TrackSize`, `Repeat`, and `LineNames`.

### DIFF
--- a/frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Grid.kt
+++ b/frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Grid.kt
@@ -94,8 +94,17 @@ sealed class GridEntry(private val value: String) {
 }
 
 private fun Array<out GridEntry>.toTrackListString(): String {
-    val entries = this@toTrackListString.also { it.validate() }
-    return entries.joinToString(" ")
+    validate()
+    return fold(mutableListOf<GridEntry>()) { acc, entry ->
+        // combine with previous line names if there is no track specified between them
+        val prev = acc.lastOrNull()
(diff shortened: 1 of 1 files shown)


bitspittle (approved):
Looks good to me! Nice refactoring.

There is one case that's broken now, but it's minor and can be resolved in a followup commit, which is specifying `lineNames` twice in a row.

```
        assertStyle("grid-template-columns: 100px [a-end b-start] 200px;") {
            gridTemplateColumns(
                GridEntry.TrackSize(100.px),
                lineNames("a-end"),
                lineNames("b-start"),
                GridEntry.TrackSize(200.px),
            )
        }
```

This produces `[a-end] [b-start]`

It could be easy to handwave this away as bad user behavior, where maybe even we just throw a runtime exception, but we at least shouldn't generate bad output. Also, I can imagine a situation where someone changes their grid based on some boolean condition, and would expect to be able to generate code like this:

```
val isPro: Boolean = checkProSubscription()

Modifier.grid {
   cols {
      lineNames("free-start")
      size(100.px)
      lineNames("free-end")
      if (isPro) {
         lineNames("pro-start")
         ...
      }
   }
}
```

Personally, I think merging the names even if declared separately is worth striving for, although I appreciate this would require pulling logic out of `LineNames` and back into `toTrackListString`. I'd be happy to take a stab at it if you just want to check in what you have right now.

bitspittle:
> Personally, I think merging the names even if declared separately is worth striving for, although I appreciate this would require pulling logic out of `LineNames` and back into `toTrackListString`. I'd be happy to take a stab at it if you just want to check in what you have right now.

Actually, probably even easier would be to do a first pass run on the track list, detect if there are any adjacent LineNames instances, remove them, and add in place a new one with them merged all together. It would probably be fairly trivial.

PR #267, opened 2023-07-06
Allow using numbers and strings as css variables

DennisTsar:


bitspittle:
Because the PR is a WIP, I can't approve it, but it honestly looks good to me so far as is. The `sealed class` approach to StyleVariables is really nice.

bitspittle (approved):

bitspittle:
You even documented the generic parameters :+1:

PR #276, opened 2023-07-18
Nit: improve stream docs wording

DennisTsar:


bitspittle:
Thanks! I like all the improvements. Also nice to have someone taking a look at my docs to verify it can be read with eyes that aren't my own.

PR #279, opened 2023-07-21
Add support for responding to stream exceptions

bitspittle:
If an API stream server endpoint crashes, this now does the following:

* Notifies the user via an onError callback
* Sends a message informing the client (with callstack when in dev mode)
* Disconnects the stream (unless the user prevents that in the onError callback)

### DIFF
--- a/backend/server/src/main/kotlin/com/varabyte/kobweb/server/plugins/Routing.kt
+++ b/backend/server/src/main/kotlin/com/varabyte/kobweb/server/plugins/Routing.kt
@@ -1,9 +1,11 @@
 package com.varabyte.kobweb.server.plugins
 
+import com.varabyte.kobweb.api.Apis
 import com.varabyte.kobweb.api.http.EMPTY_BODY
 import com.varabyte.kobweb.api.http.HttpMethod
 import com.varabyte.kobweb.api.http.Request
 import com.varabyte.kobweb.api.log.Logger
+import com.varabyte.kobweb.api.stream.ApiStream
 import com.varabyte.kobweb.api.stream.Stream
 import com.varabyte.kobweb.api.stream.StreamClientId
 import com.varabyte.kobweb.api.stream.StreamEvent
@@ -15,6 +17,7 @@ import com.varabyte.kobweb.server.api.ServerEnvironment
 import com.varabyte.kobweb.server.api.SiteLayout
 import com.varabyte.kobweb.server.io.ApiJarFile
 import com.varabyte.kobweb.streams.StreamMessage
+import com.varabyte.kobweb.streams.StreamMessage.Payload
 import io.ktor.http.*
 import io.ktor.server.application.*
 import io.ktor.server.request.*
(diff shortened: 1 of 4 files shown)


backend/server/src/main/kotlin/com/varabyte/kobweb/server/plugins/Routing.kt
```diff
@@ -39,10 +41,37 @@ import kotlin.io.path.Path
...
+private fun Throwable.stackTraceToString(includeUntil: (JavaStackTraceElement) -> Boolean): String {
+    return buildString {
+        var currThrowable: Throwable? = this@stackTraceToString
+        var lastThrowable: Throwable? = null
+        while (currThrowable != null) {
+            if (lastThrowable != null) append("caused by: ")
+            appendLine(currThrowable.toString())
+
+            // If we're handling a "caused by" stack trace, make sure the first stack trace doesn't
+            // get repeated it in.
```
DennisTsar, line 60:
```suggestion
            // get repeated in it.
```

backend/server/src/main/kotlin/com/varabyte/kobweb/server/plugins/Routing.kt
```diff
@@ -234,19 +244,45 @@ private fun Routing.setupStreaming(
...
+                                )
+                            }
                         }
+                    } catch (t: Throwable) {
+                        logger.error(
+                            "API stream (\"${incomingMessage.route}\", clientId=${id}) crashed\npayload: \"${
+                                Json.encodeToString(
+                                    incomingMessage.payload
+                                )
+                            }\"\n${t.stackTraceToString()}"
```
DennisTsar, line 271:
```suggestion
                            """
                            API stream ("${incomingMessage.route}", clientId=${id}) crashed
                            payload: "${Json.encodeToString(incomingMessage.payload)}"
                            ${t.stackTraceToString()}
                            """.trimIndent()
```

backend/server/src/main/kotlin/com/varabyte/kobweb/server/plugins/Routing.kt
```diff
@@ -234,19 +244,45 @@ private fun Routing.setupStreaming(
...
+
+                        // API streams can be created as objects or via `ApiStream` helper method. The
+                        // `includeUntil` block includes filtering logic for both cases.
+                        val callstack =
+                            if (env == ServerEnvironment.DEV) t.stackTraceToString(includeUntil = {
+                                it.className == Apis::class.qualifiedName || it.className.startsWith(
+                                    ApiStream::class.qualifiedName!!
+                                )
+                            })
+                            else null
```
DennisTsar, line 282:
```suggestion
                            if (env == ServerEnvironment.DEV) {
                                t.stackTraceToString(includeUntil = {
                                    it.className == Apis::class.qualifiedName ||
                                        it.className.startsWith(ApiStream::class.qualifiedName!!)
                                })
                            } else null
```

bitspittle:
I ended up accepting all your changes (the trimMargin one was a thing of beauty) and then committed it directly on 0.13.10 instead, so closing this PR here.

bitspittle:
Oh, also -- StreamMessage now takes a generic parameter for the payload as you suggested on Discord. Worked like a charm.

PR #300, opened 2023-08-19
Rework Switch to use checkbox input under the hood

DennisTsar:


frontend/kobweb-silk-widgets/src/jsMain/kotlin/com/varabyte/kobweb/silk/components/forms/Switch.kt
```diff
@@ -108,55 +129,66 @@ internal fun SwitchShape.toModifier() = Modifier
  *
  * @param checked Whether the switch is currently checked or not.
  * @param onCheckedChange A callback which is invoked when the switch is toggled.
- * @param modifier The modifier to apply to the *container* of this switch element. This will not be applied to the
- *   switch itself (since its configuration comes from the other parameters).
- * @param contentAlignment How to align the switch within its container. This should only be relevant if you pass in a
- *   [modifier] value that makes the size of the container larger than the switch itself. Defaults to
- *   [Alignment.CenterStart].
+ * @param modifier The modifier to apply to the *underlying checkbox input* behind this switch element. Use other
```
DennisTsar, line 132: I ultimately decided that the modifier should go to the input so that things like `Modifier.id` apply to it, which would be necessary for adding a (recommended) label.

frontend/kobweb-silk-widgets/src/jsMain/kotlin/com/varabyte/kobweb/silk/components/forms/Switch.kt
```diff
@@ -108,55 +129,66 @@ internal fun SwitchShape.toModifier() = Modifier
...
  * @param enabled Whether the switch is enabled or not. If not, the switch will be rendered in a disabled state and will
  *   not be interactable.
  * @param size The size of the switch. Defaults to [SwitchSize.MD]. You can implement your own [SwitchSize] if you want
  *   custom sizing.
  * @param colorScheme An optional color scheme to use for the switch. If not provided, the switch will use the
  *   appropriate colors from the [SilkPalette].
- * @param ref Provides a reference to the *container* of the switch. Its direct child will be the switch track, whose
- *   direct child will be the thumb element.
+ * @param focusBorderColor An optional override for the border color when the input is focused.
+ * @param ref Provides a reference to the *underlying checkbox input* of the switch. Its next sibling will be the
```
DennisTsar, line 141: This seemed to make the most sense for `ref` since it refers to the more specific 1HTMLInputElement`, but I was also considering leaving it as the container.

frontend/kobweb-silk-widgets/src/jsMain/kotlin/com/varabyte/kobweb/silk/components/forms/Switch.kt
```diff
@@ -57,6 +61,23 @@ val SwitchTrackStyle by ComponentStyle(prefix = "silk", extraModifiers = Modifie
     (hover + not(ariaDisabled)) { Modifier.cursor(Cursor.Pointer) }
 }
 
+val SwitchCheckboxVariant by InputStyle.addVariant {
+    base { // hide the checkbox itself
```
bitspittle, line 65: I would probably expand on this comment a bit. This says what the code is doing but doesn't really explain why we're hiding the checkbox.

I imagine something like, "We use an input checkbox as the source of the switch value, but then hide it visually because we render it in a different way. This helps a11y-related tools better understand and interact with this widget. This is also an approach used by many JS libraries."

frontend/kobweb-silk-widgets/src/jsMain/kotlin/com/varabyte/kobweb/silk/components/forms/Switch.kt
```diff
@@ -57,6 +61,23 @@ val SwitchTrackStyle by ComponentStyle(prefix = "silk", extraModifiers = Modifie
...
+        Modifier
```
bitspittle, line 66: Should we use visibility: hidden here instead of the weird size trick? If not, I assume it's for a11y, and probably deserves a comment?

frontend/kobweb-silk-widgets/src/jsMain/kotlin/com/varabyte/kobweb/silk/components/forms/Switch.kt
bitspittle: I'm OK with this.

frontend/kobweb-silk-widgets/src/jsMain/kotlin/com/varabyte/kobweb/silk/components/forms/Switch.kt
bitspittle: I'm really torn about this. On one hand, it makes sense that if someone wants a `ref` of a switch, they probably are thinking an element that lets them add a checked event listener, or query the `checked` status.

However, now this doesn't work: `ref = ref { it.focus() }`

which is a pretty common case too.

I believe so far that `ref` and modifier always apply to the top level container for all widgets, and we need to ask if that consistency is really worth breaking here. (It was refusing to do this earlier that helped guide the design of tabs and input groups)

I think what I'm leaning to, but note that this is definitely up for debate, is...

1) Have `modifier` apply to the container as before, but allow passing in a second `modifier` which only applies to the checkbox. This needs discussion to make sure that this meets the "Exception" case [discussed in here](https://github.com/varabyte/kobweb/blob/9eb42ecf63ab47cc5bc155dfa6dc2cb7438d3919/frontend/kobweb-silk-widgets/CONTRIBUTING.md#single-modifier-parameter). It might since here the checkbox is an ephemeral thing.

2) Create a custom scope for Switch where one of its helper methods allows you to get the ref of the inner checkbox, maybe:

```kotlin
Switch {
   checkboxRef(ref { element -> ... })
}
```

WDYT?

(Alternately, if you can find a way where `ref { it.focus() }` somehow both returns the checkbox AND works to focus the switch element, I'd be intrigued to consider it. Could Display.Contents work here?)

bitspittle (commented):
Very close to approval, just want to hash out the modifier / ref thing

bitspittle:
Just a quick followup to my review. The more I think about it, the more I feel like `ref` should return the element that represents the high level widget. You should be able to focus it and query its size and that should be meaningful. It provides a consistent mental model with also what you might see if you search the DOM for "the element" that represents the opaque representation of the whole widget.

I am more and more thinking that exposing children refs (and maybe even allowing you to set modifiers) within a special content block is a good consistent way to make special case exceptions for things like Switch, which are truly a weird widget. 

Happy to go back and forth a bit more on this. Whatever we come to, I'll enshrine in the contributing doc. However, as time passes, I am feeling stronger and stronger about ref consistency, just a heads up :)

DennisTsar:
I think the best argument for having `ref` return the checkbox is that the actual switch container is an implementation detail. However, I do like the idea of having a consistent rule. I will note that `ref { it.focus() }` does currently work as intended (since the checkbox is what applies the focus border to its sibling) and would also work if it was the container (due to how `<label>` works). 

I also like the `checkboxRef` idea. Another option would be to have Switch provide both `ref` and `checkboxRef` as params.  I do still think it should be straightforward to apply attrs (like id) to the checkbox itself, preferably with a (second?) modifier param.

bitspittle:
Weird, I swore I tried out the `it.focus` callback and it didn't work. But now it is working, so I need to think about it!

That said, I feel like if we keep to a consistent rule about what the root `modifier` and `ref` param mean, we're less likely to regret that. And consistency would imply the outer ancestor widget that contains everything else. Could be interesting to see if this is consistent with Jetpack Compose.

Honestly, I think I'm good with multiple modifier and ref params in this case. It's possible my SwitchScope block would actually not be the right abstraction.

In other words:
```
Tabs {
   TabsPanel { Tab(...); Panel(...) }
}
```
is very different conceptually from:
```
Switch {
   checkboxModifier(...)
}   
```
since the former implies a UI layout and the latter is more a complicated way to query its internals.

frontend/kobweb-silk-widgets/src/jsMain/kotlin/com/varabyte/kobweb/silk/components/forms/Switch.kt
DennisTsar: Updated. Had some context in Switch itself, but it doesn't hurt to explain here too.

DennisTsar:
I've added separate `modifier` and `ref` parameters for the checkbox. I'm open to removing `checkboxRef` though, since ultimately it's the same element as can be accessed with `it.firstChild as HTMLInputElement` from `ref`.

bitspittle (approved):

bitspittle (approved):

bitspittle:
Approved all changes. I think it's fine to have an extra ref for checkbox for now. We can see how we feel about it over time. But so far I like the modifier <-> ref explicit relationship.

Issue #307, opened 2023-09-02
Add APIs for editing SVG values


bitspittle:
For some reason, Compose HTML exposes SVG elements (like path, rect, circle) without actually declaring any of their properties.

So I guess we'll have to use extension properties and helper methods instead.

In other words, where now we have to write:

```kotlin
SVG {
    Line(attrs = {
        attr("x1", "3")
        attr("x2", "21")
        attr("y1", "12")
        attr("y2", "12")
    })
}
```
it would be way better to write:

```kotlin
// A
SVG {
   Line {
      x1(3); x2(21)
      y1(12); y2(12)
   }
}
```
or even:
```kotlin
// B
Line(x1 = 3, x2 = 21, y1 = 12, y2 = 12)
```
or even:
```kotlin
// C
Line {
   x(3, 21)
   y(12)
}
```

At a minimum, we should go through the Mozilla docs and add extension method support for all properties within an SVG element scope (case A) above, although adding useful factory methods (case B) would probably be pretty great.

binayshaw7777:
I'm interested in working on this issue.

bitspittle:
This is done and will be released in 0.14.1!

PR #314, opened 2023-09-08
Refactor SVG Element Properties for Improved Usability

binayshaw7777:
### This PR partially resolved: #307 

Added proper documentation as well!

SVGs covered so far:
- [x] Circle
- [x] Line
- [x] Rect
- [x] Polyline
- [x] Polygon
- [x] Ellipse
- [ ] Path
- [ ] Group
- [ ] Text


frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/dom/Svg.kt
```diff
@@ -53,18 +61,140 @@ fun Svg(
 
 // region SVG children
 
-// https://developer.mozilla.org/en-US/docs/Web/SVG/Element/circle
+/**
+ * A dedicated Scope for Circle
+ *
+ * Member methods: cx(), cy(), r()
```
bitspittle, line 67: So, there's a general rule about comments which is: "Don't repeat the code."

There are three good reasons for this:
1) Good code should be self documenting. If a codebase is strict about only adding comments when they are useful, then when you see one, you know it is saying something interesting. If you have a bunch of code-repeating comments, however, people start to ignore them.
2) Someone may come by in the future and change the code; it's very unlikely they'll also notice they should update the comment. It's very confusing later then when someone seems a comment that doesn't match the code and isn't sure why. This happens ALL THE TIME.
3) It's just really really depressing to work in codebases that end up looking like this:

```
/**
 * Create a foo, returning it.
 *
 * Consumes a bar.
 *
 * @param bar A bar.
 * @return foo A foo.
 */
fun createFoo(bar: Bar): Foo { ... }
```

In your comment, I don't see anything in there that I couldn't quickly figure out from looking at the code.

You can either delete the comment entirely, or, if you'd like, *maybe* use something like this:

```
/**
 * Creates a scope that provides type-safe methods for modifying an [SVGCircleElement].
 *
 * @see Circle
 */
```

This adds a little bit of information, namely it justifies the reason we're creating it (which is to wrap SVGCircleElement with a type-safe API). But it's kind of obvious looking at the code that's what it's doing, so personally I'd recommend deleting it and using nothing. (The code itself here is clear documentation)

P.S. I learned to feel very strongly about avoiding redundant comments after working with some of the best programmers in the industry, but be prepared to work for a company of mediocre developers who will ask you to do the opposite of my advice here :)

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/dom/Svg.kt
```diff
@@ -53,18 +61,140 @@ fun Svg(
...
+ *
+ * All of these above methods accept: Numbers, Length, Percentage values
+ */
+
+class SVGCircleScope(private val attrs: AttrsScope<SVGCircleElement>) : SVGElementScope(attrs) {
```
bitspittle, line 72: Let's use an internal constructor here. There's no reason a user should ever be given the opportunity to create one of these themselves:

`class SVGCircleScope internal constructor(...) `

**Please apply this comment to all relevant locations, including the base class one as well.**

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/dom/Svg.kt
```diff
@@ -53,18 +61,140 @@ fun Svg(
...
+
+    fun cx(value: CSSLengthOrPercentageValue) { attrs.attr("cx", value.value.toString()) }
+
+    fun cy(value: CSSLengthOrPercentageValue) { attrs.attr("cy", value.value.toString()) }
+
+    fun r(value: CSSLengthOrPercentageValue) { attrs.attr("r", value.value.toString()) }
+}
+
+/**
+ * Create a Circle, see example:
```
bitspittle, line 87: FYI, the first sentence of every doc comment should be a standalone sentence, as tools often use it to tag a class they're associated with.

For general guidance, see also: https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html (here, search for "First sentence")

Unlike the previous comment, where I suggested deleting it, here I might go the other way and add a bit more detail. There's a high chance that a user in the future will want to learn more about this method, unlike the scope class itself, which is more invisible to them:

```
/**
 * Type-safe API for creating an [SVGCircleElement].
 *
 * For example, to create a circle with a radius of 8 centered at 12x12:
 *
 * ```
 * Svg {
 *    Circle {
 *        ... 
 *    }
 * }
 * ```
 * See also: ...
 */
```

The fact you included an example is a nice touch. I ended up wrapping it inside an `Svg {` block to emphasize that `Circle` will only get autocompleted within that context.

**Please apply this comment to all relevant locations.**

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/dom/Svg.kt
```diff
@@ -74,34 +204,150 @@ fun ElementScope<SVGElement>.Group(
...
+ *
+ * 2. stroke() -> CSSColorValue
+ */
+class SVGPolygonScope(private val attrs: AttrsScope<SVGPolygonElement>) : SVGElementScope(attrs) {
+    fun points(vararg pairs: Pair<Number, Number>) {
+        val pointString = pairs.joinToString(" ") { "${it.first},${it.second}" }
+        attrs.attr("points", pointString)
+    }
+
+    fun stroke(value: CSSColorValue) { attrs.attr("stroke", value.toString()) }
```
bitspittle, line 233: stroke shouldn't be here; instead, it should probably be part of a common base class.

See: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke where stroke applies to pretty much everything. However, it does NOT apply to groups, so we probably want

```kotlin
abstract class SVGElementScope(...)  {
   fun attr(...)
}
abstract class SVGShapeElementScope(...) : SVGElementScope {
   fun stroke(...)
}

class SVGCircleElementScope(...) : SVGShapeElementScope { ... }
class SVGGroupElementScope(...) : SVGElementScope { ... }
```

frontend/kobweb-silk-widgets/src/jsMain/kotlin/com/varabyte/kobweb/silk/components/icons/Icons.kt
```diff
@@ -40,32 +40,36 @@ private fun createIcon(
 @Composable
 fun CheckIcon() {
     createIcon(viewBox = "0 0 12 10") {
-        Polyline(attrs = {
-            attr("points", "1.5 6 4.5 9 10.5 1")
-        })
+        Polyline {
+            points(
+                Pair(1.5, 6),
```
bitspittle, line 45: Nit: How about `points(1.5 to 6, 4.5 to 9, 10.5 to 1)`, it's a bit shorter.

frontend/kobweb-silk-widgets/src/jsMain/kotlin/com/varabyte/kobweb/silk/components/icons/Icons.kt
bitspittle, line 1: The changes in here look great!

bitspittle (commented):
Looking pretty good! 

If you've never seen it, "Nit" comments are something that was commonly used at Google at least. It means, "This is a small thing, but I wanted to mention it". It can be anything from a simple spelling error to a weakly held opinion that if you disagree with, that's fine. In general, if I ever give a review that's all nit comments, I usually just approve it at that point, so the code writer can fix what they agree with and then just submit without any further review.

Also, keep in mind, now and in the future, that a good code reviewer is not a dictator. If you disagree with something, or you have some questions about something they're asking you to do, please feel comfortable pushing back. You may still end up doing the request after a difficult back and forth, but it can be a learning process.

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/dom/Svg.kt
binayshaw7777: Right, it's a good practice for Abstraction. I didn't knew it could be done this way.

PR #315, opened 2023-09-12
Refactor SVG Element Properties for Proper and Improved Usability

binayshaw7777:
### This PR partially resolved: #307 

Added proper documentation as well!
Also resolved raised suggestions in comments and Nit comments. [Previous PR](https://github.com/varabyte/kobweb/pull/314)

SVGs covered so far:
- [x] Circle
- [x] Line
- [x] Rect
- [x] Polyline
- [x] Polygon
- [x] Ellipse
- [ ] Path
- [ ] Group
- [ ] Text

Issue #173, opened 2022-11-03
Clean up all the Modifiers that take in a string value, providing a richer API instead


bitspittle:
See :frontend:kobweb-compose `com.varabyte.kobweb.compose.ui.modifiers`

For example, `Modifier.background(value: String)` is fragile and forces users to look up the CSS APIs online. e.g. https://developer.mozilla.org/en-US/docs/Web/CSS/background

Instead, we should provide richer APIs across the board. If users want to use the text version, they can always do:

```
Modifier.styleModifier {
  background(String)
}
```

This should be done before 1.0 launches, because after that, the API is forever.

Optional: We can provide `@Deprecated` versions of these methods so users can use them if they want but get a warning.

Part of this bug is enumerating the list of all offending modifiers. That can be put into a separate comment.

PR #323, opened 2023-09-22
Refactored Text SVG

binayshaw7777:
This PR completely resolves: #307 
This PR is also a follow-up of #315 #321 

SVGs covered so far:
- [x] Circle
- [x] Line
- [x] Rect
- [x] Polyline
- [x] Polygon
- [x] Ellipse
- [x] Path
- [x] Group
- [x] Text

Attributes Covered in Text SVG:
- [x] x (Number, CSSLengthOrPercentageValue)
- [x] y (Number, CSSLengthOrPercentageValue)
- [x] dx (Number, CSSLengthOrPercentageValue)
- [x] dy (Number, CSSLengthOrPercentageValue)
- [x] rotate (vararg Number)
- [x] lengthAdjust (Enum - SvgTextLengthAdjust)
- [x] letterSpacing ((Number, CSSLengthOrPercentageValue)
- [x] wordSpacing (Number, CSSLengthOrPercentageValue)
- [x] textDecoration (TextDecorationLine)
- [x] textLength (Number, CSSLengthOrPercentageValue)
- [x] fontStyle (FontStyle)
- [x] fontSize (Number, CSSLengthOrPercentageValue, FontSize)
- [x] fontWeight (Number, FontWeight)

This also includes some implemented attributes like:
- [x] stroke (CSSColorValue)
- [x] fill (CSSColorValue, Color, Gradient)

Follow-up commit (Suggested Changes):
- Removes Fill method that takes String as Input
- Refactored values from methods that takes `CSSLengthOrPercentageValue` from `value.value.toString()` to `value.toString()`

### DIFF
--- a/frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/dom/Svg.kt
+++ b/frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/dom/Svg.kt
@@ -36,8 +36,6 @@ abstract class SVGShapeElementScope(attrs: AttrsScope<SVGElement>) : SVGElementS
 
     fun fill(value: Color) {this.attr("fill", value.toString()) }
 
-    fun fill(value: String) {this.attr("fill", value) }
-
     fun fill(value: Gradient) {this.attr("fill", value.toString()) }
 }
 
@@ -82,11 +80,11 @@ class SVGCircleScope internal constructor(private val attrs: AttrsScope<SVGCircl
 
     fun r(value: Number) { attrs.attr("r", value.toString()) }
 
-    fun cx(value: CSSLengthOrPercentageValue) { attrs.attr("cx", value.value.toString()) }
+    fun cx(value: CSSLengthOrPercentageValue) { attrs.attr("cx", value.toString()) }
 
-    fun cy(value: CSSLengthOrPercentageValue) { attrs.attr("cy", value.value.toString()) }
+    fun cy(value: CSSLengthOrPercentageValue) { attrs.attr("cy", value.toString()) }
 
-    fun r(value: CSSLengthOrPercentageValue) { attrs.attr("r", value.value.toString()) }
+    fun r(value: CSSLengthOrPercentageValue) { attrs.attr("r", value.toString()) }
 }
 
 /**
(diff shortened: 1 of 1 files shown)


frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/dom/Svg.kt
```diff
@@ -29,6 +31,14 @@ abstract class SVGElementScope(private val attrs: AttrsScope<SVGElement>) {
 
 abstract class SVGShapeElementScope(attrs: AttrsScope<SVGElement>) : SVGElementScope(attrs) {
     fun stroke(value: CSSColorValue) { this.attr("stroke", value.toString()) }
+
+    fun fill(value: CSSColorValue) {this.attr("fill", value.toString()) }
+
+    fun fill(value: Color) {this.attr("fill", value.toString()) }
+
+    fun fill(value: String) {this.attr("fill", value) }
```
bitspittle, line 39: Let's not provide "fill" which takes a string value. I'm already spending regretted time removing those I added for `Modifier` :) See also: https://github.com/varabyte/kobweb/issues/173

It's true that this will prevent some valid options, but hopefully we can identify them soon and add a typed API for them (probably a version of `fill` that takes an `SVGFill` enum value?).

In the meanwhile, a user can always use `attr("fill", "...")` for the cases we don't support yet.

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/dom/Svg.kt
```diff
@@ -431,10 +441,96 @@ fun ElementScope<SVGElement>.Rect(scope: SVGCRectScope.() -> Unit) {
...
+
+    fun dx(value: Number) { attrs.attr("dx", value.toString()) }
+
+    fun dx(value: CSSLengthOrPercentageValue) { attrs.attr("dx", value.toString()) }
+
+    fun dy(value: Number) { attrs.attr("dy", value.toString()) }
+
+    fun dy(value: CSSLengthOrPercentageValue) { attrs.attr("dy", value.toString()) }
+
+    fun rotate(vararg angleDegree: Number) {
```
bitspittle, line 473: Nit: I gave you a bad name example yesterday, sorry about that! `vararg` parameters should be plural, so here, well, `angleDegrees` isn't quite right, and `anglesDegree` isn't quite right...

I'd say let's just change this to `angles` for simplicity.

P.S. I looked at the [`sin`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.math/sin.html) method for comparison, which there is in radians, but they don't put "radians" in the parameter, so we can probably drop "degrees" here.

bitspittle (changes requested):
Looks great Binay! Essentially approved, just two minor tweaks first.

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/dom/Svg.kt
binayshaw7777: That makes sense to me, removed the overloaded method.

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/dom/Svg.kt
binayshaw7777: I understand your point. Naming convention is sure not an easy task 😅

bitspittle (approved):

PR #324, opened 2023-09-24
Fix server routing on Windows & simplify static server

DennisTsar:


bitspittle (approved):

bitspittle:
Wow love everything about this one

Issue #332, opened 2023-09-30
Support linking to other .md documents


bitspittle:
If you try to do this: `[link text](path/to/other/document.md)`, currently that generates a link that fails (because in the final site, document.md will have been converted into document.html by that time). It would be nice of a target .md link was detected and converted into a URL automatically.

Note that this is supported in GitHub pages - you can target an md document and the link will get automatically converted to an html url.

There are two reasons we should support this feature, besides the fact that GH is doing it...

1. Kobweb markdown allows route overrides ([see README](https://github.com/varabyte/kobweb#route-override-1)), so it would be pretty cool to link to an md file knowing that whatever final URL it resolves to will be handled automatically.(*)

2. When navigating markdown files in GitHub, it can follow links to other markdown files in the same project.

---

(*) Example of routeOverride:

```
# TargetPage.md
---
routeOverride: target-page
---

# SourcePage.md

Link to [target page][TargetPage.md]
```

should generate final Kotlin code that looks like

```kotlin
Text("Link to ")
Link("target-page", "target page")
```

---

> [!NOTE]
> I set this bug to "good first issue" because it's relatively self contained, but there is a chance that the devil is in the details with solving the `routeOverride` stuff, so it may be harder than I'm expecting.

bitspittle:
(This feature actually went out in 0.14.1)

PR #333, opened 2023-10-01
Replace links to .md files in Markdown with their generated counterparts

MartinTheDragon:
Implements the feature request in #332, works correctly with route overrides.
Doesn't support linking to files with dynamic route overrides, and will currently simply emit a warning and proceed normally. Please tell if an error and/or stopping execution is more desirable.

### DIFF
--- a/tools/gradle-plugins/extensions/markdown/src/main/kotlin/com/varabyte/kobwebx/gradle/markdown/KotlinRenderer.kt
+++ b/tools/gradle-plugins/extensions/markdown/src/main/kotlin/com/varabyte/kobwebx/gradle/markdown/KotlinRenderer.kt
@@ -346,7 +346,7 @@ class KotlinRenderer(
         override fun visit(link: Link) {
             // Links to other Markdown files, if the files are present, get converted to their corresponding generated routes
             if (link.destination.endsWith(".md")) {
-                val destinationPath = Path(filePath.substringBeforeLast('/', "")).resolve(link.destination).normalize().toString()
+                val destinationPath = Path(filePath).resolveSibling(link.destination).normalize().toString()
                 val destinationFile = markdownNodeGetter(destinationPath.removePrefix("/"))
                 if (destinationFile != null) {
                     // Retrieve the destination's route override, if present
--- a/tools/gradle-plugins/extensions/markdown/src/main/kotlin/com/varabyte/kobwebx/gradle/markdown/KotlinRenderer.kt
+++ b/tools/gradle-plugins/extensions/markdown/src/main/kotlin/com/varabyte/kobwebx/gradle/markdown/KotlinRenderer.kt
@@ -44,6 +44,8 @@ import org.commonmark.renderer.Renderer
 import org.gradle.api.Project
 import org.gradle.api.provider.Provider
 import java.util.*
+import kotlin.io.path.Path
+import kotlin.io.path.relativeToOrSelf
 
 fun String.yamlStringToKotlinString(): String {
     return if (this.isSurrounded("\"")) {
@@ -57,6 +59,7 @@ fun String.yamlStringToKotlinString(): String {
 
 class KotlinRenderer(
     private val project: Project,
+    private val markdownNodeGetter: (path: String) -> Node?,
     private val imports: List<String>,
     private val filePath: String,
     private val handlers: MarkdownHandlers,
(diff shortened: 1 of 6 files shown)


tools/gradle-plugins/extensions/markdown/src/main/kotlin/com/varabyte/kobwebx/gradle/markdown/KotlinRenderer.kt
```diff
@@ -340,6 +344,49 @@ class KotlinRenderer(
         }
 
         override fun visit(link: Link) {
+            // Links to other Markdown files, if the files are present, get converted to their corresponding generated routes
+            if (link.destination.endsWith(".md")) {
+                val destinationPath = Path(filePath.substringBeforeLast('/', "")).resolve(link.destination).normalize().toString()
+                val destinationFile = markdownNodeGetter(destinationPath.removePrefix("/"))
```
bitspittle, line 350: Optional: Maybe "destinationNode"

tools/gradle-plugins/extensions/markdown/src/main/kotlin/com/varabyte/kobwebx/gradle/markdown/KotlinRenderer.kt
```diff
@@ -340,6 +344,49 @@ class KotlinRenderer(
...
+                    // Retrieve the destination's route override, if present
+                    val frontMatterData = with(FrontMatterVisitor()) {
+                        destinationFile.accept(this)
+                        this.data
+                    }
+                    val routeOverride = getRouteOverride(destinationPath, frontMatterData)
+
+                    if (routeOverride != null) {
+                        if ('{' in routeOverride)
+                            reporter.warn("Markdown file link '${link.destination}' links to file with dynamic route override. This is not supported!")
```
bitspittle, line 361: This is a great check. Should this be an if/else as well?
```kotlin
if ('{' in routeOverride) { 
  warn
} else { // valid route
   val prefix = ...
}
```

tools/gradle-plugins/extensions/markdown/src/main/kotlin/com/varabyte/kobwebx/gradle/markdown/KotlinRenderer.kt
```diff
@@ -340,6 +344,49 @@ class KotlinRenderer(
...
+
+                        val prefix = if (routeOverride.startsWith('/')) {
```
bitspittle, line 363: No action required in this commit. Just leaving a note for myself because I'm wondering if I should be putting some common logic somewhere, since I feel like it must exist in multiple places now.

bitspittle (approved):
Beautiful! Just wanted to say, getting a PR out of the blue like this made my day. Thanks for taking the time to figure out the code and the project organization.

I think your NodeCache refactoring is great; I appreciate the simplicity of cleanliness of the approach. Also, naming is good -- just enough information to be clear.

I left 3 very trivial comments (actually one is for myself, not for you, but you're welcome to weigh in on it if you had any opinions since you're seeing this code with fresh eyes). 

But really, I'm happy to accept this change as is (squashed with the followup change to protect against escaping the markdown folder, which I'll review next).

tools/gradle-plugins/extensions/markdown/src/main/kotlin/com/varabyte/kobwebx/gradle/markdown/tasks/ConvertMarkdownTask.kt
```diff
@@ -132,14 +133,18 @@ abstract class ConvertMarkdownTask @Inject constructor(
     private class NodeCache(private val parser: Parser, private val roots: List<File>) {
         private val existingNodes = mutableMapOf<String, Node>()
 
-        fun get(file: File): Node = existingNodes.computeIfAbsent(file.toUnixSeparators()) {
+        fun get(file: File): Node = existingNodes.computeIfAbsent(file.canonicalFile.toUnixSeparators()) {
             parser.parse(file.readText())
         }
 
-        fun getRelative(relPath: String): Node? =
+        fun getRelative(relPath: String): Node? = try {
```
bitspittle, line 140: Nit: Let's add a short header comment here, since the return condition is a little more nuanced now

Feel free to use this or your own wording if you prefer:

/**
 * Returns a parsed Markdown node given a relative path which will be resolved against all cache roots.
 *
 * For example, "test/example.md" will return parsed markdown information if found in `src/jsMain/resources/markdown/test/example.md`.
 *
 * This will return null if:
 * * no file is found matching the passed in path.
 * * the file at the specified location fails to parse.
 * * the relative file path escapes the current root, e.g. `../public/files/license.md`, as this could be a useful way 
 *    to link to a raw markdown file that should be served as is and not converted into an html page.
 */

bitspittle (approved):

tools/gradle-plugins/extensions/markdown/src/main/kotlin/com/varabyte/kobwebx/gradle/markdown/KotlinRenderer.kt
MartinTheDragon: Yeah, that's what I was asking about in the PR message. I chose to leave it at a warning and proceeding normally because a filename *may* contain curly braces, and I didn't want to make it impossible to link to files that have those in their name but wouldn't actually be recognized as dynamic routes.
Maybe there should be a more sophisticated check via Regex than just looking for a curly brace anywhere in the string.

PR #338, opened 2023-10-05
Updated README.md

ScalarJerk:
Made changes to enhance readability

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,14 +1,21 @@
-![version: 0.14.1](https://img.shields.io/badge/kobweb-0.14.1-blue)
-![version: 0.9.13](https://img.shields.io/badge/kobweb_cli-0.9.13-blue)
+<p align="center">
+<img src="https://github.com/varabyte/media/raw/main/kobweb/screencasts/kobweb-welcome.gif" />
+</p>
+<p align="center">
+<img alt="Version" src="https://img.shields.io/badge/kobweb-0.14.1-blue">
+<img alt="CLI Version" src="https://img.shields.io/badge/kobweb_cli-0.9.13-blue">
+<img alt="Kotlin Version" src="https://img.shields.io/badge/kotlin-1.9.10-blue?logo=kotlin">
 <br>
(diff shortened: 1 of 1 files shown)


bitspittle:
Thank you for the PR. I appreciate the idea behind making sure Kobweb's README's first impression is stronger. After getting it, I actually reached out to a designer friend to ask them for some thoughts as well, and checked out some other READMEs.

For example...
* Kotlin: https://github.com/JetBrains/kotlin#readme
* nextjs: https://github.com/vercel/next.js#readme
* shields.io: https://github.com/badges/shields#readme
* langchain: https://github.com/langchain-ai/langchain#readme

It seems like left-aligned and centered badges are both used, although I think the aesthetic of a centered logo with centered badges does look a bit nicer.

However, the solution here isn't to move the animated demo image to the top. That's intentionally kept close to the code -- the idea is that developers in this case are my main audience, so code is the most important thing. They should see the code and the result it generates side by side. I think it's very important to keep those two concepts connected.

Note that on my website https://kobweb.varabyte.com, you'll see I reversed the order, because that looks a bit better. However, I couldn't get it to look good on GitHub (because in that case it's an image, not a resizable dynamic element), so I left it below the code in this case. 

Finally, the changes you made to NOTE and IMPORTANT sections are actually incorrect. The way is was written before is the official way to create note / warning / important sections on GitHub. You can read more here: https://github.com/orgs/community/discussions/16925

You can see this for yourself!

Before your change:
note-before

After your change:
note-after

---

**So, in its current state, I hope you can understand why I'm not going to accept this PR.** However, I do think if I could make a good large copy of the Kobweb logo (I need to reach out to the guy who made the smaller version), that putting that at the top and centering the badges under it might look a bit better.

Also, in the next few months (hopefully!) there's going to be a huge push to move a lot of the README out to the Kobweb website (linked above), at which point we can control the aesthetics a lot more than we can on GitHub.

ScalarJerk:
Should I revert the Note, Warning and Important to what they were? I didn't notice the before state while I was updating them locally. My apologies

PR #342, opened 2023-10-06
Add AlwaysUnderlined link variant

### LINKED ISSUE #57: Add an "AlwaysUnderlined" link variant
Right now, the default behavior is to only underline if the cursor moves over the link. Some people might want always underlined behavior, and while it's easy enough to let them write it, it's also trivial for us to provide a variant for them.

Geanik:
Closes #57

### DIFF
--- a/frontend/silk-widgets-kobweb/src/jsMain/kotlin/com/varabyte/kobweb/silk/components/navigation/Link.kt
+++ b/frontend/silk-widgets-kobweb/src/jsMain/kotlin/com/varabyte/kobweb/silk/components/navigation/Link.kt
@@ -58,6 +58,12 @@ val UncoloredLinkVariant by LinkStyle.addVariant {
     visited { colorModifier }
 }
 
+val AlwaysUnderlinedLinkVariant by LinkStyle.addVariant {
+    base {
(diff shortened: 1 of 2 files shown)


bitspittle:
Thank you so much for the PR.

Fun fact -- as you can tell, I could have easily done this myself, but I left this issue around because I felt it would be a good test to see if I'd ever get third party contributors. This was still pretty early into Kobweb, so that wasn't a given. I then promptly forgot about it.

I appreciate you being the dev who finally stepped up and knocked this bug out. :raised_hands:

bitspittle:
Ah so I thought it would be easy for me to adjust the base branch for you, but apparently GitHub's UI is getting confused by that. I could just pull your change down locally and fix it myself on the command line, but I worry that will mess up ownership of this PR.

**So please rebase your change against 0.14.2, and when that's done, I'll merge it.** You can [read more about this here](https://github.com/varabyte/kobweb/blob/main/CONTRIBUTING.md#latest-branch). Thanks!

Geanik:
should be fine now right?
changed the base branch to 0.14.2.

thx for keeping this issue for new devs.
was a nice task for having a bit of a look into your repo 👍

bitspittle:
Yep, perfect. Thanks!

PR #351, opened 2023-10-14
Preliminary implementation of dispose callback generation for server api

zypus:
This pull request accompanies this discussion #350 

It adds a @DisposeApi annotation next to the @InitApi annotation and adds the encessary code generation to add a dispose function to the ApiJarFile. 

This dispose function is currently called when the cache of the ApiJarFileImpl gets invalidated. This currently is only relevant for the DEV mode.

As the ApiJarFileImpl is quite deeply hidden and I already exposed it from the cache, I did not implement that the dispose function is also called when the actual server shutdown, but it should for consistency be also called on shutdown. Hence this part would still be a TODO.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,6 @@
+blank_issues_enabled: false
+contact_links:
+  - name: Ask a question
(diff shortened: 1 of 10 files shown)


zypus:
Did branch from main which was incorrect, gonna fix that and open a new pull request.

PR #352, opened 2023-10-15
Dispose api callback, cleanup api after dev server reload

zypus:
This pull request accompanies this discussion https://github.com/varabyte/kobweb/discussions/350

It adds a @DisposeApi annotation next to the @InitApi annotation and adds the encessary code generation to add a dispose function to the ApiJarFile.

This dispose function is currently called when the cache of the ApiJarFileImpl gets invalidated. This currently is only relevant for the DEV mode.

As the ApiJarFileImpl is quite deeply hidden and I already exposed it from the cache, I did not implement that the dispose function is also called when the actual server shutdown, but it should for consistency be also called on shutdown. Hence this part would still be a TODO.


zypus:
Closing this pull request in favor of #353

Issue #254, opened 2023-06-07
Expose a way that users can use Ktor auth


bitspittle:
For context, see: https://ktor.io/docs/authentication.html

To get auth in ktor to work, you need to install authentication and then wrap various routes with "authentication" blocks. However, Kobweb sets up the routes for you.

So the question is...

1) Is this still a relevant feature for something like Kobweb where you download the whole site at once and do mostly client-side navigation?

2) Would this be better handled using a Kobweb server plugin? Or by writing a kobwebx library that people can use to manage logged in states?

bitspittle:
A possible approach for Kobweb server plugins would be to add a method that looks something like this:

```
fun wrapRoute(path: String, block: () -> Unit): (() -> Unit)?
```

By default, this would return null, but you might be able to add something like this:

```
return if (path.startsWith("/protected")) {
   return {
      basic_auth { content() }
   }
} else null
```

bitspittle:
It's possible that this feature would benefit from "Kobweb Spidering", the tentative name I have for separating server components apart (so that if you link inside your site to `example.com/admin`, it will forcefully download new content instead of navigating internally in the JS already pulled down)

bitspittle:
Could be worth seeing how Android Compose apps handle auth

mrlem:
Interesting topic indeed.

I'm quite new to Kobweb, so sorry if I don't know all its features, but if I may intervene, here's what usually happens on android apps (at least those I worked on):
* you call an authentication API which provides you with some token
* this token is being passed in all subsequent API calls
* if at any point you get a 401 in an API call while being logged-in, have an HTTP interceptor spots and reports this as a session loss

Usually, all this is buried in lower layers (data layer in a clean architecture). But these usually provide a flow about authentication status based on what I described above, which can then be passed on to the upper layers to the view-state, which is sent to composables.

I see no reason why such a flow could not work with Kobweb. The way I see it:
* **client side**, there's no need for an `@Auth` annotation or `requireAuth`: because this would cover one case (checking whether the user is authenticated or not), but not others (e.g having permissions / profiles and such).
* **server side**, we indeed need a way to identity the user. Say I want to perform some checks in @Apis in order to reject the user

Taking the example of [session authentication](https://ktor.io/docs/server-session-auth.html), which relies on cookies, what you need is basically:
* server (ktor server already offers that):
  - be able to send cookies to the client
  - be able to store data associated with each cookie 
* client:
  - just send requests with your site cookies (which is automatic)

[...]

PR #353, opened 2023-10-15
Implements api events, currently only the dispose event is present, but can be extended in the future

zypus:
Implements the events and dispose callback feature as discusses in #350.

Modifies the server code and the gradle plugin such that the following is now possible:

```kotlin
@InitApi
fun initDatabase(ctx: InitApiContext) {
    val store = // create database

    ctx.events.onDispose { event: DisposeEvent ->
        store.close()
    }

    // setup database
}
```

The InitApiContext gets an additional `events` member. This currently allows to register callbacks for dispose events via the `onDispose` method.

Currently a dispose event is triggered in the following cases:

1. During DEV mode the api.jar is reloaded, will call `dispose(DisposeEvent(reason=DisposeReason.DEV_API_RELOAD))`
2. When the server shuts down normally, will call `dispose(DisposeEvent(reason=DisposeReason.SERVER_SHUTDOWN))`
3. In case the server is shutdown, because the JVM itself is shutdown/killed, will call `dispose(DisposeEvent(reason=DisposeReason.SHUTDOWN_HOOK))` 

 @bitspittle The third option registers a shutdownHook on the engine, maybe that would also be interesting to move some of the current server shutdown code into, because if the server is shutdown unnaturally some of the current shutdown code will not be run.


backend/kobweb-api/src/main/kotlin/com/varabyte/kobweb/api/event/Events.kt
```diff
@@ -1,21 +1,25 @@
 package com.varabyte.kobweb.api.event
 
-import com.varabyte.kobweb.api.dispose.DisposeApiContext
+import com.varabyte.kobweb.api.dispose.DisposeEvent
 
 interface Events {
-    fun onDispose(callback: (ctx: DisposeApiContext) -> Unit)
+    fun onDispose(callback: (event: DisposeEvent) -> Unit)
```
bitspittle, line 6: This is probably the right place to add a comment about "dispose" not being guaranteed.

I'll write one here not because I want you to take it blindly, but because I used to hate back and forths with code reviewers because they wanted me to write a comment but weren't clear about what they wanted me to say.

Use this as is or adjust it to what you think is best, up to you! I'm going to write it out quickly without thinking too much about what I'm writing:

```
/**
 * Registers a handler which will be triggered when unmanaged server resources should be disposed.
 *
 * A server might initialize some resources that should be removed explicitly if informed that the server
 * is coming to the end of its lifecycle. If so, this callback is the place to handle this.
 *
 * It's important to understand there may be circumstances where this method can't be called, such as
 * a program being forcefully terminated or the machine it is running on losing power. In other words, even
 * though this event is provided, you have to assume there's a chance it won't get called, and you must write
 * your server logic in a way that can recover from these events.
 *
 * That said, this can be very useful at development time, as live reloading may cause a server to restart
 * multiple times in a short period of time.
 */
```

backend/kobweb-api/src/main/kotlin/com/varabyte/kobweb/api/dispose/DisposeEvent.kt
```diff
@@ -0,0 +1,12 @@
+package com.varabyte.kobweb.api.dispose
+
+/**
+ * An enum that represents the reason for the api disposal
+ */
+enum class DisposeReason {
+    DEV_API_RELOAD, SERVER_SHUTDOWN, SHUTDOWN_HOOK
```
bitspittle, line 7: Let's add some short comments to these? Maybe:

- A dispose is requested in response to a dev server being restarted due to a live reload request
- A dispose is requested because the server is about to shut down normally
- A dispose is requested because the JVM running the server was requested to shut down, e.g. in response to Ctrl-C

bitspittle (changes requested):
Honestly, this is pretty great! You not only got the chance working the way I imagined it should, but you were even very careful to pay attention to the rest of the project (as far as I can tell) when it came to things like parameter order, naming, etc.

Thank you so much!

The change is functionally fine as is, with some very minor comments that should be easy to resolve.

backend/kobweb-api/src/main/kotlin/com/varabyte/kobweb/api/event/Events.kt
zypus: Thanks a lot, that is indeed very helpful to get an idea what is required. I restructured and reworded your example a bit and expanded on suggesting potential use cases.

zypus:
Thank you. This change was quite straightforward due to the clear and clean structure you have build. I am glad that I managed to implement it close to what you have envisioned. I am also really happy how clean it ended up for the user side.

I am glad I judged the structure and order of parameters correctly. I am well aware of the importance of names and structure :D.

Happy to contribute, I really like what you are building with kobweb.

bitspittle (approved):

bitspittle:
> Happy to contribute, I really like what you are building with kobweb.

Thanks! Really happy to hear that. And thanks for making Kobweb better even with this change! Changes of your quality level are welcome anytime.

If you haven't seen it yet, we have a general roadmap for what we feel is missing for a 1.0 release:
https://docs.google.com/document/d/1n2Jd02yzuxaatpT7gOhEuijzfaSu9UIaQTV2t4EPcPk/preview

That said, we're getting to a point where a lot of what remains in pretty mind numbing stuff. :) You can review that and see if anything in general interests you. 

Since I don't think that many people are using full stack at this point, if you just keep using Kobweb and pushing that feature to its limits, you might come up with a bunch more feature requests that would actually benefit a 1.0 release that I'm totally blind to.

bitspittle:
Heads up that I just kicked off a new snapshot. It should be ready whenever this comment is at least 20 minutes old.

At that point, you should be able to use your changes if you change your Kobweb version to `0.14.2-SNAPSHOT`

zypus:
You're welcome and thank you the discussions about the feature. 

I am glad I could add something to the project. I will check out the roadmap and I will probably stumble upon some more features that would be nice to have. Probably the next thing I might have to deal with is authentication related.

bitspittle:
As we don't have too many backend folks yet, I think if you just do backend work in your project and stay in touch with me if you have ideas, well, that would be great!

Keep in mind auth in a website produced by a framework like Kobweb is tricky, because you download the whole site at once. You can contrast this to using ktor directly, where you could add auth to routes on a case by case basis, because the user has to hit the network each time to change pages.

My notes are a little sloppy here, but you can take a look at #254 to see a few more details of what I was thinking. 

In Kobweb, I suspect auth will be best served by 1) integration on the frontend side and/or 2) an awesome auth library on the backend (plus some manual labor to check auth in each relevant API request)

## frontend


Despite the user downloading the whole site, we could still have logic to redirect to a different page if credentials are not found.

It's possible auth is handled by some `@Auth` annotation that I haven't really thought too deeply about yet.

## backend

With a backend approach, I think you'd let users navigate to any page, but on each page that requires auth, you'd make an API call first to check if they're logged in, and if not, reroute to a login page then. (You can use local storage or session storage to cache a login key).

And then, don't ever return data from any of your protected API routes if the user doesn't pass in identifying keys in the get params or post body.

It's fuzzy in my memory now, but I think I do something like this a bit with `kobweb create examples/chat`

---

Of course, open to hear any other ideas you have! Or if you find any other feature missing from the backend that you think would be interesting to work on.

PR #369, opened 2023-10-24
Limit visibility of mutable properties

DennisTsar:


frontend/silk-foundation/src/jsMain/kotlin/com/varabyte/kobweb/silk/components/style/StyleModifiers.kt
```diff
@@ -14,7 +14,7 @@ import org.jetbrains.compose.web.css.*
  */
 abstract class StyleModifiers {
     private val _cssModifiers = mutableListOf<CssModifier>()
-    internal val cssModifiers: List<CssModifier> = _cssModifiers
+    internal val cssModifiers: List<CssModifier> get() = _cssModifiers
```
bitspittle, line 17: You shouldn't need `get()` here because `_cssModifiers` never gets reassigned?

bitspittle (approved):

frontend/silk-foundation/src/jsMain/kotlin/com/varabyte/kobweb/silk/components/style/StyleModifiers.kt
bitspittle: (Same comment applies to SilkTheme.kt changes as well)

DennisTsar: You're right, fixed.

PR #372, opened 2023-11-06
`ComponentStyle` simplification

DennisTsar:


frontend/silk-foundation/src/jsMain/kotlin/com/varabyte/kobweb/silk/components/style/ComponentStyle.kt
```diff
@@ -167,11 +133,11 @@ class ComponentStyle(
         }
     }
 
-    internal fun addStylesInto(styleSheet: StyleSheet, selectorName: String) {
+    internal fun addStylesInto(styleSheet: StyleSheet, selectorName: String): List<String> {
```
bitspittle, line 208: What do you think of creating an internal class called `StyleSelectors` which is just an inline class that wraps a List called `parts` or `classnames` (we can move the `substringAfterLast` utility method in there and add a `toString` method which is ".${parts.joinToString(".")}"?

The reason it's nice to have a class is it will be a good place to add header docs and make the code in this commit look less generic -- WDYT?

bitspittle (approved):

bitspittle (approved):

frontend/silk-foundation/src/jsMain/kotlin/com/varabyte/kobweb/silk/components/style/ComponentStyle.kt
DennisTsar: Good idea, implemented. I ended up leaving most of the docs on the functions themselves, since it's relevant to usage itself, but open to feedback on that.

bitspittle (approved):

PR #376, opened 2023-11-17
First steps of improving the docs

ellet0:
Just a few changes to the current docs

Like fixing typos, making it easier to understand, adding more info, and splitting the sections into different pages
which makes the height of the page much less which result into making it much easier to maintain and read

Yes this doc or `README.md` is far from perfect
But if you want my honest and simple opinion, any `README.md` would be preferred over the current one :)

Feel free to merge or reject the changes


.idea/kotlinc.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="KotlinJpsPluginSettings">
-    <option name="version" value="1.9.10" />
+    <option name="version" value="1.9.20" />
```
freshtechtips, line 4: I didn't update this, my IDE did

.idea/misc.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="ExternalStorageConfigurationManager" enabled="true" />
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" project-jdk-name="jdk-11" project-jdk-type="JavaSDK">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" project-jdk-name="corretto-17" project-jdk-type="JavaSDK">
```
freshtechtips, line 4: Same for this, even though I see 17 is more preferred over 11

But shouldn't this file ignored in `.gitignore`??

bitspittle:
Hey!!

I really, really appreciate the effort here.

I totally agree with you that the super long README is not great. Not only is it hard to navigate, there are sections missing that I'm hesitant to add because it would makes the whole thing even more unwieldy.

I will take some time to think about your PR. However, just so you know that the only reason I would be hesitant to take it is because the goal has always been to move the docs to my docs site (https://kobweb.varabyte.com/docs), so I don't want to take too much time right now to tweak the README since I'm going to be doing that soon with the docs site. It's not because I disagree with your goal to organize things into pieces. 

I've been postponing the docs migration not because I'm lazy and hate docs, but because I wanted to finalize some decisions about the widget APIs (we did a ton of work on this in the last few months). This was to prevent me writing a bunch of docs on my site, only to have to rewrite them again shortly after. Right now, there's one foundational change left to the Kobweb APIs that we're working on, and once that's done, it will be time to migrate the docs. This is currently looking like a few more weeks.

**When I start migrating docs to the site, would you like to be involved?**

bitspittle:
Also: If you make a change where you just tweak readability / fix typos without doing the reorganization of files, those I could accept a lot easier right now.

.idea/misc.xml
bitspittle: I was using the standard gitignore template from JB. Can you find a link saying that `.idea/kotlinc.xml` and `.idea/misc.xml` should be ignored?

bitspittle: Also, keep in mind that what Java version to use as an app developer is _different_ than what a library author should consider. Apps should use the latest versions while library authors should use the oldest versions they can for maximum compatibility.

freshtechtips:
> Also: If you make a change where you just tweak readability / fix typos without doing the reorganization of files, those I could accept a lot easier right now.

I don't know if you saw the new documentation in an editor or preview it's really not that much, so it's easy to consider if you want to merge or reject

It's easier to see the changes in a IDE preview than git

bitspittle:
> Do you have Slack Channel?? I prefer over Discord

There's a #kobweb channel on the official Kotlin slack, you can find me there.

bitspittle:
> I don't know if you saw the new documentation in an editor or preview it's really not that much, so it's easy to consider if you want to merge or reject

I'm very busy today so it will be a day or two before I can look. The more work I have to do the longer it will take :) Typo fixes are trivial to accept, rewording is potentially harder as I'll need to think about it / see if I have any of my own suggestions, and full on doc reorganization is going to be harder still, because then I need to think about what I want to be doing in the future and if this approach fits.

Doc reorganization isn't an automatic win either like you might think it is, because sometimes people (myself for example when answering questions on my Discord) go to the main Kobweb README and search for keywords to find sections to link to. Once doc pages are nested, that will be significantly harder to do. When I migrate the docs to the Kobweb site I will have a search bar which should only search docs pages (making it much easier to use than GitHub's frustrating search feature). I hope this explanation makes sense!

Meanwhile, please review https://github.com/varabyte/kobweb/blob/main/CONTRIBUTING.md#kobweb-development which talks about how PRs should be done against the latest branch, not `main` (at the moment, `0.15.1`, but I will be releasing in a day probably so it might change to `0.15.2` soon in case you want to wait). I suspect if you work out of the latest dev branch, then you won't get the weird tool file changes, as I probably checked in fixes for those on the latest branch. (I will probably be reviewing this at some point in the future now that we're getting more people contributing, but that's the way it will be for at least a little longer)

.idea/misc.xml
freshtechtips: > Also, keep in mind that what Java version to use as an app developer is _opposite_ to what a library author should consider. Apps should use the latest versions while library authors should use the oldest versions they can for maximum compatibility.

Yes, right, but we also need to test it with Java JDK 17

The same applies to all dependencies versions

But that's another subject

bitspittle:
Closing this PR due to the above discussions. Further work is gated on docs moving over to the website, which hopefully will happen early-ish next year.

PR #398, opened 2023-12-20
Add caption-side css style and Kobweb modifiers

ellet0:
Like the title

This PR might needs to more review and suggestions in order to complete


frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/CaptionSide.kt
```diff
@@ -0,0 +1,36 @@
+package com.varabyte.kobweb.compose.css
+
+import org.jetbrains.compose.web.css.*
+
+fun StyleScope.captionSide(value: String) {
```
bitspittle, line 5: We've been making the decision not to support [stringly-typed](https://wiki.c2.com/?StringlyTyped) APIs at this point. Any time a CSS property takes a String in the Kobweb codebase means that property actually accepts arbitrary string values OR maybe it's a legacy mistake. If you see any others that aren't marked deprecated, let me know.

For now, let's delete this.

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/CaptionSide.kt
```diff
@@ -0,0 +1,36 @@
...
+        val BlockStart get() = CaptionSide("block-start")
+        val BlockEnd get() = CaptionSide("block-end")
+        val InlineStart get() = CaptionSide("inline-start")
+        val InlineEnd get() = CaptionSide("inline-end")
+
+        // Global values
+        val Inherit get() = CaptionSide("inherit")
+        val Initial get() = CaptionSide("initial")
+        val Revert get() = CaptionSide("revert")
+        val RevertLayer get() = CaptionSide("revert-layer")
```
bitspittle, line 29: Ah, this is subtle and I apologize for forgetting to mention it.

Compose HTML at the moment doesn't support [CSS layers](https://developer.mozilla.org/en-US/docs/Web/CSS/@layer), so we don't define the RevertLayer global value (even though you putting it here was the technically correct thing to do).

> [!NOTE]
> I used to define `RevertLayer` myself but noticed the Compose HTML APIs never did, so eventually I decided to join them (since at the moment there's never a reason for a user to use it).

frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/ui/modifiers/CaptionSideModifiers.kt
```diff
@@ -0,0 +1,13 @@
+package com.varabyte.kobweb.compose.ui.modifiers
+
+import com.varabyte.kobweb.compose.css.*
+import com.varabyte.kobweb.compose.ui.Modifier
+import com.varabyte.kobweb.compose.ui.styleModifier
+
+fun Modifier.captionSide(value: String) : Modifier = styleModifier {
```
bitspittle, line 7: Delete stringly-typed API here

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/CaptionSide.kt
bitspittle, line 1: Our file names in this package are semi- abitrary, but we try to choose general categories that we can fit multiple CSS properties in. One CSS property per Kotlin file would explode :)

Here, I'd suggest calling this `Table.kt` (and `TableModifiers.kt` later), since when I read the docs, it mentions how this is a CSS property that affects tables.

bitspittle (changes requested):
The definition of the CaptionSide class is nearly perfect, basically exactly as I imagined it (the only thing to tweak in it actually was my fault, not yours, but you can read more in the comment).

Besides that, there are two broad issues with the review that prevent me from accepting it --

1. It wants to merge against main. Please update that to the 0.15.4 branch.
2. I'm not sure why it keeps happening, but I keep seeing kotlinc.xml and misc.xml changes leaking into your reviews. I wonder if maybe we can do a video chat at some point so I can watch how you are creating your commits? If you're not comfortable with that, we can chat over Kotlin slack to figure it out. If I was uploading this PR, it would only have a single commit in it, and the commit would just be the two Kotlin files (one in `compose-html-ext` and one in `kobweb-compose`).

ellet0:
> 2\. I'm not sure why it keeps happening, but I keep seeing kotlinc.xml and misc.xml changes leaking into your reviews. I wonder if maybe we can do a video chat at some point so I can watch how you are creating your commits? If you're not comfortable with that, we can chat over Kotlin slack to figure it out. If I was uploading this PR, it would only have a single commit in it, and the commit would just be the two Kotlin files (one in `compose-html-ext` and one in `kobweb-compose`).

Sure thing

ellet0:
> (the only thing to tweak in it actually was my fault, not yours, but you can read more in the comment).

No-one has fault, we are here to understand each other code.

ellet0:
> 2. I'm not sure why it keeps happening, but I keep seeing kotlinc.xml and misc.xml changes leaking into your reviews. I wonder if maybe we can do a video chat at some point so I can watch how you are creating your commits? If you're not comfortable with that, we can chat over Kotlin slack to figure it out. If I was uploading this PR, it would only have a single commit in it, and the commit would just be the two Kotlin files (one in `compose-html-ext` and one in `kobweb-compose`).

Once I open the project using IntelliJ IDEA it will change those files, I copied the original files to revert the changes but the IDE still auto format the files for unknown reasons no matter what I do, I'm sure I can fix it but should we format xml code of `.idea` files anyway?

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/CaptionSide.kt
ellet0: I commented it instead of deleting it, in case we go back in the future we can revert this more quickly in case someone is not familiar with Css properties, and for second we are trying to implement what the docs of [mozilla](developer.mozilla.org/en-US/docs/Web/CSS/caption-side) said, even if we don't plan on using RevertLayer

let me know if I should remove the comment of `RevertLayer`

frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/ui/modifiers/CaptionSideModifiers.kt
ellet0: Done

ellet0:
I'm unable to fix the conflicts or getting it in my local machine

bitspittle:
> Once I open the project using IntelliJ IDEA it will change those files, I copied the original files to revert the changes but the IDE still auto format the files for unknown reasons no matter what I do, I'm sure I can fix it but should we format xml code of .idea files anyway?

What I'd recommend doing is just keeping those changes around locally and not committing them. Only commit and push the changes that should be reviewed in other words.

I don't want to make any assumptions about your git level, so if you feel I'm speaking down to you, let me know! I'm just being cautious....

Are you comfortable adding files individually into a commit and leaving other files in a modified state? Do you have experience with `git amend`? How about `git apply` to apply patches?

ellet0:
> What I'd recommend doing is just keeping those changes around locally and not committing them. Only commit and push the changes that should be reviewed in other words.
> 
> I don't want to make any assumptions about your git level, so if you feel I'm speaking down to you, let me know! I'm just being cautious....
> 
> Are you comfortable adding files individually into a commit and leaving other files in a modified state? Do you have experience with `git amend`? How about `git apply` to apply patches?

So to make it more simple, you want me to squash all the commits into one commit?

ellet0:
> > Once I open the project using IntelliJ IDEA it will change those files, I copied the original files to revert the changes but the IDE still auto format the files for unknown reasons no matter what I do, I'm sure I can fix it but should we format xml code of .idea files anyway?
> 
> What I'd recommend doing is just keeping those changes around locally and not committing them. Only commit and push the changes that should be reviewed in other words.
> 
> I don't want to make any assumptions about your git level, so if you feel I'm speaking down to you, let me know! I'm just being cautious....
> 
> Are you comfortable adding files individually into a commit and leaving other files in a modified state? Do you have experience with `git amend`? How about `git apply` to apply patches?

I get it now, just woke up so I didn't get it you right

ellet0:
I don't know why but I keep getting templates module as changed even though I didn't touch it, besides things got a little missy after I switch to the `0.15.4` branch to merge, this [new PR ](https://github.com/varabyte/kobweb/pull/399/files)

PR #400, opened 2023-12-21
Add appearance css property and modifier

ellet0:


frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Element.kt
```diff
@@ -65,3 +65,33 @@ fun StyleScope.content(value: String) {
 
 fun CSSUrl.toContent() = Content.of(this)
 fun Gradient.toContent() = Content.of(this)
+
+
+// https://developer.mozilla.org/en-US/docs/Web/CSS/appearance
+class Appearance private constructor(private val value: String) : StylePropertyValue {
```
bitspittle, line 71: Nit: I'm not always perfect about this myself, but let's alphabetize here instead of putting at the end. (I was going to mention this with your other PR about the `clear` property but I noticed that file was somewhat arbitrarily ordered, so I'm planning to reorder it after your change goes in)

My thoughts are that ordering of code should be intentional. This does not always mean alphabetical -- you can group things into clumps of related classes when that makes more sense.

However, alphabetical is better than just default dropping stuff at the end of the file. The reason for this is, if it becomes a habit and everyone on a team dumps things at the end, then the chance of an unnecessary merge conflict goes up and up.

bitspittle (approved):
I think I gave you my first "Nit" comment here. "Nit"s are things which I have an opinion about but don't feel extremely strongly about / wouldn't block a PR over. If a review is all nits, I tend to approve them.

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Element.kt
bitspittle: BTW after making this comment, I went ahead and cleaned up my code around the `Layout.kt` properties and alphabetized things: https://github.com/varabyte/kobweb/commit/2a066fa8f0984a79e11060bb2996b92fe9bf63b9

Thanks for helping me be less lazy!

bitspittle:
I'm going to go ahead and accept this change and tweak the ordering on my end, it's so trivial. Otherwise I think it can be easy to accidentally miss picking up this PR before the next release.

PR #401, opened 2023-12-21
Add build.yml workflow

ellet0:
I noticed when we make typos or build errors the code, we can still send the PR and there is no CI that check the code build, so I added this simple GitHub workflow called `build.yml` to help us with this

It will work on all PRs and only when push to the main branch.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,31 @@
+name: Build
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ '*' ]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: gradle/wrapper-validation-action@v1
+
+      - name: Setup JDK 11
+        uses: actions/setup-java@v3
+
+        with:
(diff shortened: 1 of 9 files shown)


ellet0:
See? we can get the benefit of this PR already, we got caught a test failure in this PR

.github/workflows/build.yml
```diff
@@ -0,0 +1,31 @@
+name: Build
+
+on:
+  push:
```
bitspittle, line 4: Let's just do this on pull_requests for now.

I'd like to review it in that context first and then we can extend to push actions later.

bitspittle (changes requested):
Looks good, thanks!

ellet0:
> test-all.yml

I think both your name and mine doesn't fit but it doesn't much matter because we don't have much workflows, just two until now, I will rename it

.github/workflows/build.yml
ellet0: > Let's just do this on pull_requests for now.
> 
> I'd like to review it in that context first and then we can extend to push actions later.

Done

ellet0:
> Should the file here be called "test-all.yml" or something instead of "build.yml"?

Renamed

PR #402, opened 2023-12-21
Replace the markdown template with GitHub new form templates

ellet0:
This one is simple, it replace the markdown templates (bug report and feature request) with the new GitHub forms to provide the data in a better way, input and output
it looks better and give the user better experience

I copied the old text from the markdown and paste them into the new GitHub forms so no need for them anymore
I already test them on another repo I have

Also added question new issue template to make it clear that's a question and not a bug or feature request


.github/ISSUE_TEMPLATE/1_bug.yml
```diff
@@ -0,0 +1,123 @@
...
+    attributes:
+      value: |
+        Thank you for using Kobweb!
+
+  - type: checkboxes
+    attributes:
+      label: Is there an existing issue for this?
+      options:
+        - label: I have searched the [existing issues](https://github.com/varabyte/kobweb/issues)
+          required: true
```
bitspittle, line 17: required: false, default to checked true. I am even on the fence about removing this one entirely.

It might be fine to have this in as a reminder, but I don't want people to have to click a checkbox just to communicate with me. 

We can revisit this if bug volume ever becomes a problem, but so far, it isn't, and people have already been good about searching the bug database themselves before filing to be honest. This just hasn't been a problem.

.github/ISSUE_TEMPLATE/1_bug.yml
```diff
@@ -0,0 +1,123 @@
...
+      description: Please tell us which version of `Kobweb` that you are using.
+      placeholder: For example 0.15.4
+    validations:
+      required: true
+  - type: textarea
+    attributes:
+      label: Other versions
+      description: If relevant, specify the Kobweb CLI version, Kobweb libs version and the Operating system
+      placeholder: |
+        * OS: _(Choose one: Linux/Mac/Win)_
```
bitspittle, line 30: Hmm, I don't like this text as a placeholder. I want this text to be something that's already written that the user can either tweak in place or delete. If it's a placeholder, once they start typing parts of it, then the placeholder will disappear and they'll forget the rest.

I'm guessing this should be separating into three different inputs. One for Kobweb version (which you have above), one for CLI version, and one for OS (which should be a pulldown)

.github/ISSUE_TEMPLATE/1_bug.yml
```diff
@@ -0,0 +1,123 @@
...
+    attributes:
+      label: Is there an existing issue for this?
+      options:
+        - label: I have searched the [existing issues](https://github.com/varabyte/kobweb/issues)
+          required: true
+  - type: input
+    attributes:
+      label: Kobweb version
+      description: Please tell us which version of `Kobweb` that you are using.
+      placeholder: For example 0.15.4
```
bitspittle, line 22: placeholder: from your `gradle/libs.version.toml` file

.github/ISSUE_TEMPLATE/1_bug.yml
```diff
@@ -0,0 +1,123 @@
...
+    validations:
+      required: true
```
bitspittle, line 24: required: false

People can add the version if they want but it's usually not relevant. If I try to repro and I can't, I can just ask them anyway.

.github/ISSUE_TEMPLATE/1_bug.yml
```diff
@@ -0,0 +1,123 @@
...
+      description: Please tell us what is expected to happen.
+    validations:
+      required: true
+  - type: textarea
+    attributes:
+      label: Actual results
+      description: Please tell us what is actually happening.
+    validations:
+      required: true
+  - type: textarea
```
bitspittle, line 57: Let's delete this section. Code should go in the "describe the issue" section.

.github/ISSUE_TEMPLATE/1_bug.yml
```diff
@@ -0,0 +1,123 @@
...
+        <details><summary>Code sample</summary>
+
+        ```dart
+        [Paste your code here]
+        ```
+
+        </details>
+    validations:
+      required: true
+  - type: textarea
```
bitspittle, line 81: Let's delete. People can just put images directly into their bug report.

.github/ISSUE_TEMPLATE/1_bug.yml
```diff
@@ -0,0 +1,123 @@
...
+      value: |
+        <details>
+        <summary>Screenshots / Video demonstration</summary>
+
+        [Upload media here]
+
+        </details>
+    validations:
+      required: false
+  - type: textarea
```
bitspittle, line 95: Also this should probably just be in the bug description.

I'm potentially willing to leave this in because of the extra information in the description, but honestly, this hasn't ever been a problem once so far, so I'm not sure it's worth the noise.

.github/ISSUE_TEMPLATE/3_question.yml
```diff
@@ -0,0 +1,24 @@
+name: Ask a question
```
bitspittle, line 1: Questions shouldn't be asked in the issue tracker. They should instead go to the discussion tab.

bitspittle (commented):
Hey, I appreciate what you're trying to do here, especially I wasn't familiar with the new form templates and it's neat to see them in action.

That said, I think you'll notice in my comments that are already a lot just for the bug template suggestion (and there were others I didn't make yet). I think there's way more thought that needs to go into this -- it's not a straightforward conversion like it might seem.

Ultimately, this feels like we're spending both of our time now trying to solve a problem that's not actually a problem (and in doing so risk introducing new, real problems). 

(In the future, feel free to ping me on Slack to ask about actual pain points users or I am having on this project so we can focus on solving those first; or if you are OK with getting this feedback over code reviews, that's fine too, but I'm hoping to save your time).

For now, I believe the best course of action is to reject this PR and revisit this in the future if bug management ever starts to get overwhelming / users start spamming us with low quality bugs.

PR #407, opened 2023-12-22
Add css accent color attribute and modifier

ellet0:


bitspittle (changes requested):
Looks good.

Did you see my comment in my other PR about sort ordering stuff instead of just appending to the end of the file?

(I would suggest here as well sorting AccentColor higher up in the file)

bitspittle (approved):

bitspittle (approved):

PR #411, opened 2024-01-02
CSS `Appearance` improvements

DennisTsar:


frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Element.kt
```diff
@@ -13,7 +13,7 @@ class Appearance private constructor(private val value: String) : StylePropertyV
         // CSS Basic User Interface Module Level 4 values
         val None get() = Appearance("none")
         val Auto get() = Appearance("auto")
-        val MenuListButton get() = Appearance("menulist-button")
+        val MenulistButton get() = Appearance("menulist-button")
```
dhcoder, line 16: I feel like "Menulist" looks better as "MenuList"?

Note that "TextField" (left unchanged below) would technically be "Textfield" as well if we followed the same logic (but "Textfield" also looks wrong to me because it's "TextField" in so many UI APIs.

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Element.kt
```diff
@@ -13,7 +13,7 @@ class Appearance private constructor(private val value: String) : StylePropertyV
         // CSS Basic User Interface Module Level 4 values
         val None get() = Appearance("none")
         val Auto get() = Appearance("auto")
-        val MenuListButton get() = Appearance("menulist-button")
```
bitspittle, line 16: I feel like "Menulist" looks better as "MenuList"?

Note that "TextField" (left unchanged below) would technically be "Textfield" as well if we followed the same logic (but "Textfield" also looks wrong to me because it's "TextField" in so many UI APIs.

DennisTsar: I think that it's better to be consistent with css when we can, and you're right, that means using "Textfield" as well. We have made exceptions ("NoWrap", though I prefer "Nowrap"), but since this is an uncommon property I don't think it's worth making an exception for it. That being said, I'm open to keeping the names as they are.

PR #412, opened 2024-01-02
Add support for corner-specific and elliptical border radii

### LINKED ISSUE #251: Add support for elliptical border radii
https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius note the "slash" versions.

DennisTsar:
Fixes #251


bitspittle (approved):
Looks fantastic.

Not related to this PR, but do you think in the future are you against me adding "startStart", "startEnd", "endStart", and "endEnd" methods into BorderRadiusScope as well? It's a slight deviation from CSS spec but it seems like the most sensical place to add them when we support those versions?

PR #414, opened 2024-01-02
Add css attribute multi-column

ellet0:


frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/MultiColumn.kt
```diff
@@ -0,0 +1,51 @@
...
+    override fun toString() = value
+
+    companion object {
+
+        // Global values
+        val Inherit get() = MultiColumn("inherit")
+        val Initial get() = MultiColumn("initial")
+        val Revert get() = MultiColumn("revert")
+
+        //        val RevertLayer get() = MultiColumn("revert-layer")
```
bitspittle, line 16: I've seen this before from a different commit, so I thought I'd mention that this  commented out `RevertLayer` line shouldn't be included at all. Instead, just have Inherit, Initial, Revert, and Unset be all next to each other.

FYI this is eventually going to get refactored out into a separate interface anyway, so it's going to go away eventually :) But in general, you shouldn't ever submit commented out code EXCEPT in comments that are explaining code examples for clarity (as far as I can think of right now)

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/MultiColumn.kt
```diff
@@ -0,0 +1,51 @@
+package com.varabyte.kobweb.compose.css
+
+import org.jetbrains.compose.web.css.*
+
+// https://developer.mozilla.org/en-US/docs/Web/CSS/column-rule
+class MultiColumn private constructor(private val value: String) : StylePropertyValue {
```
bitspittle, line 6: I think there may have been some confusion in what I was saying.

I meant the _file_ should be called "MultiColumn.kt", but the name of each class should come from its CSS property.

Here, that means we'll have classes / stylescope methods here for:

```
break-after
break-before
break-inside
column-fill
column-gap
column-span
column-rule-color
column-rule-style
column-rule-width
column-count
column-width
```

I'm not sure each CSS property needs a class (honestly, `column-rule-color`, `style`, and `width` can probably just be functions).

But if you DO need to create a class (for example, https://developer.mozilla.org/en-US/docs/Web/CSS/break-after will need one), then that would be `class BreakAfter`).

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/MultiColumn.kt
```diff
@@ -0,0 +1,51 @@
+package com.varabyte.kobweb.compose.css
+
+import org.jetbrains.compose.web.css.*
+
+// https://developer.mozilla.org/en-US/docs/Web/CSS/column-rule
```
bitspittle, line 5: Note that "column-rule" is a shorthand property for "column-rule-color", "column-rule-width", and "column-rule-style". So we shouldn't define a `ColumnRule` class.

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/MultiColumn.kt
```diff
@@ -0,0 +1,51 @@
...
+        val Inherit get() = MultiColumn("inherit")
+        val Initial get() = MultiColumn("initial")
+        val Revert get() = MultiColumn("revert")
+
+        //        val RevertLayer get() = MultiColumn("revert-layer")
+        val Unset get() = MultiColumn("unset")
+    }
+}
+
+fun StyleScope.multiColumn(multiColumn: MultiColumn) {
```
bitspittle, line 21: This method shouldn't exist at all

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/MultiColumn.kt
```diff
@@ -0,0 +1,51 @@
...
+    property("column-rule", multiColumn)
+}
+
+fun StyleScope.multiColumn(color: CSSColorValue) {
```
bitspittle, line 25: This should be `StyleScope.columnRuleColor`

This comment applies to the methods below, as well as to the methods defined in MultiColumnModifiers.kt

bitspittle (changes requested):

frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/ui/modifiers/MultiColumnModifiers.kt
```diff
@@ -0,0 +1,32 @@
+package com.varabyte.kobweb.compose.ui.modifiers
+
+import com.varabyte.kobweb.compose.css.*
+import com.varabyte.kobweb.compose.ui.Modifier
+import com.varabyte.kobweb.compose.ui.styleModifier
+import org.jetbrains.compose.web.css.*
+
+fun Modifier.multiColumn(multiColumn: MultiColumn) = styleModifier {
```
bitspittle, line 8: Please check what "BorderModifiers.kt" is doing, with its introduction of a `BorderScope` class. We'll want to copy that style over here.

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/MultiColumn.kt
ellet0: Done

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/MultiColumn.kt
```diff
@@ -0,0 +1,49 @@
...
+        val Revert get() = ColumnRule("revert")
+        val Unset get() = ColumnRule("unset")
+    }
+}
+
+fun StyleScope.columnRule(columnRule: ColumnRule) {
+    property("column-rule", columnRule)
+}
+
+fun StyleScope.columnRuleColor(color: CSSColorValue) {
```
bitspittle, line 23: Everything here and down is perfect.

bitspittle (changes requested):

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/MultiColumn.kt
```diff
@@ -0,0 +1,49 @@
...
+
+        // Global values
+        val Inherit get() = ColumnRule("inherit")
+        val Initial get() = ColumnRule("initial")
+        val Revert get() = ColumnRule("revert")
+        val Unset get() = ColumnRule("unset")
+    }
+}
+
+fun StyleScope.columnRule(columnRule: ColumnRule) {
```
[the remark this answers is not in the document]
ellet0: So we will not support the global keywords?

frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/ui/modifiers/MultiColumnModifiers.kt
```diff
@@ -0,0 +1,32 @@
+package com.varabyte.kobweb.compose.ui.modifiers
+
+import com.varabyte.kobweb.compose.css.*
+import com.varabyte.kobweb.compose.ui.Modifier
+import com.varabyte.kobweb.compose.ui.styleModifier
+import org.jetbrains.compose.web.css.*
+
+fun Modifier.columnRule(columnRule: ColumnRule) = styleModifier {
```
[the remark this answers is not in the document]
ellet0: How about now? I think we still do need more changes

frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/ui/modifiers/MultiColumnModifiers.kt
```diff
@@ -0,0 +1,37 @@
+package com.varabyte.kobweb.compose.ui.modifiers
+
+import com.varabyte.kobweb.compose.css.*
+import com.varabyte.kobweb.compose.ui.Modifier
+import com.varabyte.kobweb.compose.ui.graphics.Colors
+import com.varabyte.kobweb.compose.ui.styleModifier
+import org.jetbrains.compose.web.css.*
+class ColumnRuleScope internal constructor(private val styleScope: StyleScope) {
```
bitspittle, line 8: Nit: Space between imports and class. (Actually, can you just run `Code > Reformat Code`?)

frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/ui/modifiers/MultiColumnModifiers.kt
```diff
@@ -0,0 +1,37 @@
...
+    fun color(color: CSSColorValue) = styleScope.borderColor(color)
+    fun style(lineStyle: LineStyle) = styleScope.borderStyle(lineStyle)
+    fun width(width: CSSLengthNumericValue) = styleScope.borderWidth(width)
+    fun width(topAndBottom: CSSLengthNumericValue, leftAndRight: CSSLengthNumericValue) =
```
bitspittle, line 12: No action required: I'm actually talking with another dev on Kobweb and in just the last few days we decided that we should be calling this "topBottom" and "leftRight", without the "And".

If you submit this as is, that's fine, I can fix it shortly when we make the change, but if you're going to do the "reformat code" anyway, would you mind renaming this?

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/MultiColumn.kt
```diff
@@ -0,0 +1,31 @@
+package com.varabyte.kobweb.compose.css
+
+import org.jetbrains.compose.web.css.*
+
```
bitspittle, line 4: Optional: What do you think about a comment at the top of the file linking to Mozilla docs, like:

```
package com.varabyte.kobweb.compose.css

// This file contains CSS properties as described here:
// https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_multicol_layout

import org.jetbrains.compose.web.css.*

fun StyleScope.columnRuleColor(color: CSSColorValue) {
```

bitspittle (approved):
Approved with super minor requests. Thank you for this PR!

bitspittle:
Merged via command line, guess I need to close the PR manually! 
https://github.com/varabyte/kobweb/commit/30e3c34c06cfff789dc445811f90909320847350

PR #417, opened 2024-01-03
CSS Scroll Improvements

DennisTsar:


frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Scroll.kt
```diff
@@ -24,18 +24,37 @@ fun StyleScope.scrollBehavior(scrollBehavior: ScrollBehavior) {
 
 // region Scroll snap
 // See https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-snap-type
-class ScrollSnapType private constructor(private val value: String) : StylePropertyValue {
+sealed class ScrollSnapType private constructor(private val value: String) : StylePropertyValue {
     override fun toString() = value
 
+    private class Keyword(value: String) : ScrollSnapType(value)
+    class Axis internal constructor(value: String) : ScrollSnapType(value)
+    private class WithStrictness(axis: Axis, strictness: Strictness) : ScrollSnapType("$axis $strictness")
```
bitspittle, line 32: Totally optional: `AxisWithStrictness`

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Scroll.kt
```diff
@@ -57,15 +85,25 @@ class ScrollSnapMode private constructor(private val value: String) : StylePrope
 
     companion object {
         // Keyword
+        @Deprecated(
+            "Use `ScrollSnapType.Strictness.Mandatory` instead.",
```
bitspittle, line 89: Stale comments I think. `of(axis, Mandatory)` instead

bitspittle (approved):
Only nits

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Scroll.kt
DennisTsar: The idea behind the deprecations is that a previous use of `Modifier.scrollSnapType(ScrollSnapAxis.Both, ScrollSnapMode.Mandatory)` ends up with a deprecation `ReplaceWith` for each parameter, which after being applied will become the valid `Modifier.scrollSnapType(ScrollSnapType.Both, ScrollSnapType.Strictness.Mandatory)`

PR #420, opened 2024-01-04
docs: Update README.md

MahmoudMabrok:
remove `$ ` as when command being it be copied also, which is wrong

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ need to do this in your own project!*
 *OS: Mac and Linux*
 
 ```bash
-$ brew install varabyte/tap/kobweb
+ brew install varabyte/tap/kobweb
 ```
 
 ### [Scoop](https://scoop.sh/)
@@ -125,7 +125,7 @@ $ brew install varabyte/tap/kobweb
 *OS: Windows, Mac, and \*nix*
 
 ```shell
-$ sdk install kobweb
+ sdk install kobweb
 ```
 
 ### Don't see your favorite package manager?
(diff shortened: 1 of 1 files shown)


bitspittle:
Hey -- thank you for uploading a draft PR that we can discuss.

I'm on the fence here and I need to think about this a bit more. I did check other repos, and it was illuminating -- in every case I found, they presented their bash commands without things like a "$" so they are easy to copy.

However, on the flip side, I designed many of my examples as something to read, not to copy. In fact, copying most of my examples would be very bad :) I'll show you what I mean:

---

```bash
$ cd /path/to/applications

# You can either pull down the zip file

$ wget https://github.com/varabyte/kobweb-cli/releases/download/v0.9.13/kobweb-0.9.13.zip
$ unzip kobweb-0.9.13.zip

# ... or the tar file

$ wget https://github.com/varabyte/kobweb-cli/releases/download/v0.9.13/kobweb-0.9.13.tar
$ tar -xvf kobweb-0.9.13.tar
```
^ This is a choice the user has to make. If they run both, it's going to end up with one command clobbering the other. Also "/path/to/applications" isn't a real path. It will be a different value for each user.

---

```bash
$ PATH=$PATH:/path/to/applications/kobweb-0.9.13/bin
$ kobweb version # to check it's working
```
^ This is a hint that users should update the path however they're comfortable doing so. Usually this means editing your `.bashrc` file but explaining all the ways to do this was outside the scope of the README.

---

```bash
$ cd /path/to/bin # some folder you've created that's in your PATH
$ ln -s /path/to/applications/kobweb-0.9.13/bin/kobweb kobweb
```
^ "/path/to/bin" does not exist

---

```bash
JAVA_HOME=/path/to/jdks/corretto-11.0.12
# ... or whatever version or path you chose
```
^ Again, the user has to make a choice. Copying directly is almost certainly wrong for most users.

---

```bash
$ cd /path/to/src/root # some folder you've created for storing src code
$ git clone https://github.com/varabyte/kobweb-cli
$ cd kobweb-cli
$ ./gradlew :kobweb:installDist
```
^ "/path/to/src/root does not exist

---

```bash
$ kobweb list

You can create the following Kobweb projects by typing `kobweb create ...`

• app: A template for a minimal site that demonstrates the basic features of Kobweb
• examples/jb/counter: A very minimal site with just a counter (based on the Jetbrains tutorial)
• examples/todo: An example TODO app, showcasing client / server interactions
```
^ This is demonstrating output that a user sees on a terminal when they run `kobweb list`, it's not something to copy.

---

**Let me assure you, I can see that removing the "$" could be useful in some situations!**

The way things are now, I think we'd need to approach this more carefully and intentionally than just removing the dollar signs. The right way forward here might be to rethink how I shared examples in the first place.

Also note that at some point many of these README bits are going to move over to a docs site, and I don't think the docs site will include copy buttons like github does.

So at the moment, I hope you can understand my hesitation. I'm not against changing my examples here, especially around installing kobweb via brew and other package managers. But a simple find/replace "$" -> "" isn't the way.

MahmoudMabrok:
totally agree.

PR #436, opened 2024-01-18
Remove unnecessary functions in `KobwebModuleTask`

DennisTsar:
We do not need to depend on the buildscripts directly, as each task declares the inputs it needs, and gradle will rerun the task if they change.


bitspittle (approved):
I swear when I added this code in it made a difference, but maybe things were different two years ago, or maybe I'm just wrong!

Either way, happy to delete code. Thanks!

PR #447, opened 2024-01-26
Use a more idiomatic approach for copying the server jar

DennisTsar:


bitspittle (approved):
Neat!

I've learned more about gradle watching you than all the docs in the world combined

PR #450, opened 2024-01-28
Fix KSP using incomplete file dependencies

DennisTsar:
This fixes an issue introduced by #444 where not-yet-complete file dependencies were being passed into KSP's codeGenerator, resulting in incomplete output upon recompilation. Now, the data and files are returned by a single function call, ensuring that the latter cannot be accessed prematurely.

### DIFF
--- a/tools/ksp/site-processors/src/main/kotlin/com/varabyte/kobweb/ksp/frontend/AppProcessor.kt
+++ b/tools/ksp/site-processors/src/main/kotlin/com/varabyte/kobweb/ksp/frontend/AppProcessor.kt
@@ -50,9 +50,9 @@ class AppProcessor(
     }
 
     override fun finish() {
-        val (frontendData, files) = frontendProcessor.getFinalizedData()
-        fileDependencies.addAll(files)
-        val appData = AppData(appFqn?.let { AppEntry(it) }, frontendData)
+        val frontendResult = frontendProcessor.getProcessorResult()
+        fileDependencies.addAll(frontendResult.fileDependencies)
+        val appData = AppData(appFqn?.let { AppEntry(it) }, frontendResult.data)
 
         val (path, extension) = genFile.split('.')
         codeGenerator.createNewFileByPath(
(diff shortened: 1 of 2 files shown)


tools/ksp/site-processors/src/main/kotlin/com/varabyte/kobweb/ksp/frontend/AppProcessor.kt
```diff
@@ -50,8 +50,9 @@ class AppProcessor(
     }
 
     override fun finish() {
-        fileDependencies.addAll(frontendProcessor.fileDependencies)
-        val appData = AppData(appFqn?.let { AppEntry(it) }, frontendProcessor.getData())
+        val (frontEndData, files) = frontendProcessor.getFinalizedData()
```
bitspittle, line 53: Nit: frontendData

bitspittle, line 53: How come we don't have to do this with the backendProcessor?

tools/ksp/site-processors/src/main/kotlin/com/varabyte/kobweb/ksp/frontend/FrontendProcessor.kt
```diff
@@ -189,16 +189,26 @@ class FrontendProcessor(
...
+    /**
+     * Get the finalized metadata acquired over all rounds of processing.
+     *
+     * This function should only be called from [SymbolProcessor.finish] as it relies on all rounds of processing being
+     * complete.
+     *
+     * @return A pair consisting of the finalized FrontendData and a list of file dependencies that should be passed in
+     * when using KSP's [CodeGenerator] to store the data.
+     */
+    fun getFinalizedData(): Pair<FrontendData, List<KSFile>> {
```
bitspittle, line 211: It's not the worst that this is a Pair, since the code is isolated to a processor. That said, I'd be good introducing a `data class FinalizedData(val frontendData, val fileDependencies)` and using that here. Your calling code shouldn't have to change.

bitspittle (approved):

tools/ksp/site-processors/src/main/kotlin/com/varabyte/kobweb/ksp/frontend/AppProcessor.kt
DennisTsar: BackendProcessor does not rely on an additional processor, so just uses its own file dependencies. But AppProcessor  needs to get data and dependencies from the internal FrontendProcessor that it uses.

bitspittle (approved):

PR #451, opened 2024-01-29
Fix worker extension arguments being resolved eagerly

DennisTsar:


bitspittle (approved):
Might be worth linking to this commit to the KSP folks, in case we can demonstrate how we'd simplify this code if they made some changes. (I assume this is related to what we want them to allow us to do?)

PR #453, opened 2024-01-31
Connect markdown task input/outputs

DennisTsar:


tools/gradle-plugins/extensions/markdown/src/main/kotlin/com/varabyte/kobwebx/gradle/markdown/tasks/ProcessMarkdownTask.kt
```diff
@@ -72,29 +64,32 @@ abstract class ProcessMarkdownTask @Inject constructor(
     )
 
     @OutputDirectory
-    fun getGenResDir(): File = kobwebBlock.getGenJsResRoot<MarkdownBlock>(project).resolve(
-        project.prefixQualifiedPackage(kobwebBlock.publicPath.get()).replace(".", "/")
```
bitspittle, line 76: Oh good catch, thanks!

bitspittle (approved):
This overall looks like a great approach, certainly gradle idiomatic.

On one hand I guess part of me does wonder if it's going to be harder for new users to follow what's going on, since they can't get all the context directly inside the MarkdownConvert and MarkdownProcess tasks. On the other, the way you hook up task inputs / outputs in KobwebxMarkdownPlugin feels good.

Let me know if there was anything about the design you were still thinking through.

PR #454, opened 2024-01-31
Ensure gradle property helper functions are only called once

DennisTsar:
This should prevent users from mistakenly overriding values they already set.


bitspittle (approved):
Very interesting. Thanks!

PR #461, opened 2024-02-04
Specify roles of custom gradle configurations

DennisTsar:
The default behavior, where `isCanBeConsumed` and `isCanBeResolved` are both true, is deprecated.


bitspittle:
Looks like it's working! Merging.

PR #463, opened 2024-02-06
Add a Gradle task which lists all routes defined by the current site

bitspittle:
While doing this, refactor some code out of "export" into a separate task so that both tasks can share code.


DennisTsar (approved):

bitspittle:
Just to make sure -- you're also OK with the random path I chose for the output file? "build/kobweb/cache/frontendData.json"?

DennisTsar:
I was fine with the path choice, but you're right that if we intend to use the "cache" directory for other tasks then it would be have to be adjusted to prevent overlapping outputs.

bitspittle:
Cool, in a followup commit, I'll update the convention to be "kobweb/cache/$taskName", maybe even via a utility method.

I'll be honest, the idea of dropping temporary files as outputs that then get parsed as inputs in other tasks is really neat. I suppose this is Gradle-idiomatic but it never really occurred to me that it was a possibility until now. Instead I was trying to hard to constrain myself to serializable java objects.

PR #510, opened 2024-04-21
Add the modifiers `widthIn`, `heightIn`, and `sizeIn`, and reformat "SizeModifiers.kt"

ShreckYe:
I'd like to propose these modifiers that are more akin to those in Compose in addition to the `min...`/`max...` ones.

### DIFF
--- a/frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/ui/modifiers/SizeModifiers.kt
+++ b/frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/ui/modifiers/SizeModifiers.kt
@@ -23,25 +23,25 @@ fun Modifier.size(size: CSSLengthOrPercentageNumericValue): Modifier = size(widt
 
 fun Modifier.size(width: CSSLengthOrPercentageNumericValue, height: CSSLengthOrPercentageNumericValue): Modifier =
     styleModifier {
-    width(width)
-    height(height)
-}
+        width(width)
+        height(height)
+    }
 
 fun Modifier.minSize(size: CSSLengthOrPercentageNumericValue): Modifier = minSize(width = size, height = size)
 
 fun Modifier.minSize(width: CSSLengthOrPercentageNumericValue, height: CSSLengthOrPercentageNumericValue): Modifier =
     styleModifier {
-    minWidth(width)
-    minHeight(height)
-}
+        minWidth(width)
+        minHeight(height)
+    }
 
 fun Modifier.maxSize(size: CSSLengthOrPercentageNumericValue): Modifier = maxSize(width = size, height = size)
 
 fun Modifier.maxSize(width: CSSLengthOrPercentageNumericValue, height: CSSLengthOrPercentageNumericValue): Modifier =
     styleModifier {
-    maxWidth(width)
-    maxHeight(height)
-}
+        maxWidth(width)
+        maxHeight(height)
+    }
 
 fun Modifier.width(size: CSSLengthOrPercentageNumericValue): Modifier = styleModifier {
     width(size)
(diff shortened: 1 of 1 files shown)


frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/ui/modifiers/SizeModifiers.kt
```diff
@@ -98,3 +98,28 @@ fun Modifier.maxHeight(size: CSSLengthOrPercentageNumericValue): Modifier = styl
...
+    styleModifier {
+        minWidth(min)
+        maxWidth(max)
+    }
+
+fun Modifier.widthIn(min: MinWidth, max: MaxWidth): Modifier = styleModifier {
+    minWidth(min)
+    maxWidth(max)
+}
+
```
bitspittle, line 112: // Mimics Jetpack Compose API https://developer.android.com/reference/kotlin/androidx/compose/foundation/layout/package-summary#(androidx.compose.ui.Modifier).heightIn(androidx.compose.ui.unit.Dp,androidx.compose.ui.unit.Dp)

frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/ui/modifiers/SizeModifiers.kt
```diff
@@ -98,3 +98,28 @@ fun Modifier.maxHeight(size: CSSLengthOrPercentageNumericValue): Modifier = styl
...
+        minHeight(min)
+        maxHeight(max)
+    }
+
+fun Modifier.heightIn(min: MinHeight, max: MaxHeight): Modifier = styleModifier {
+    minHeight(min)
+    maxHeight(max)
+}
+
+fun Modifier.sizeIn(min: CSSLengthOrPercentageNumericValue, max: CSSLengthOrPercentageNumericValue): Modifier =
```
bitspittle, line 124: https://developer.android.com/reference/kotlin/androidx/compose/foundation/layout/package-summary#(androidx.compose.ui.Modifier).sizeIn(androidx.compose.ui.unit.Dp,androidx.compose.ui.unit.Dp)

bitspittle, line 124: This looks wrong.

According to

https://developer.android.com/reference/kotlin/androidx/compose/foundation/layout/package-summary#(androidx.compose.ui.Modifier).sizeIn(androidx.compose.ui.unit.Dp,androidx.compose.ui.unit.Dp,androidx.compose.ui.unit.Dp,androidx.compose.ui.unit.Dp)

this should have four parameters.

bitspittle (commented):
Hey, thank you for your three PRs (and I appreciate you merging against the latest dev branch, thanks for reading the docs!).

Since this is the first time you've worked with me, some general notes:

* You are allowed to disagree / push back. I am always willing to entertain debate to get to the best final result. That said, if we get stuck on something, it's best if we can provide concrete reasons to each other or links to articles or docs in the wild to support our way of thinking. Also, sometimes if two people get stuck on an approach, it could be a sign that a third approach exists out that we should try to discover.

* I use "Optional" and "Nit" in my comments sometimes. "Optional" means I'm just throwing something out there because I had a thought but not a very strong one, and I really don't care if you just dismiss the comment. "Nit" is a little stronger than that -- I'd still be fine if a "Nit" change doesn't go in, but I think there's some value in the comment. If a review is 100% Optional and Nit comments, I'll usually just approve it.

* Sometimes, I write really long, detailed comments, including code snippets. I am not doing this as a way to bludgeon the other person into submission to accept my way. I'm just trying to be as clear as possible of what I'm thinking about, with the hopes that this reduces the number of back and forths needed to get a PR in. I used to hate dealing with code reviewers where someone pushed back against my changes without telling me exactly what they wanted, requiring multiple emails to get to the bottom of their concerns. 

I am sure our goals are the same -- 1) get the code review done with the minimum number of back and forths and 2) get the best solution we can think of in.

Thanks again!
David

frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/ui/modifiers/SizeModifiers.kt
ShreckYe: You are right. I made a mistake here.

ShreckYe:
Hi @bitspittle, thank you for your reply. I have applied your suggestions.

bitspittle (approved):
Looks great, thanks!

PR #511, opened 2024-04-21
Add one more overloaded `size` modifier

ShreckYe:


frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/ui/modifiers/SizeModifiers.kt
```diff
@@ -27,6 +27,9 @@ fun Modifier.size(width: CSSLengthOrPercentageNumericValue, height: CSSLengthOrP
     height(height)
 }
 
+fun Modifier.size(width: Width, height: Height): Modifier =
```
bitspittle, line 30: As discussed in the other PR, I don't think we should support keyword versions of these methods.

ShreckYe: No problem.

PR #512, opened 2024-04-21
Add the `wrapContent...` modifiers

ShreckYe:
I'd like to propose these modifiers that are more akin to those in Compose.

### DIFF
--- a/frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/ui/modifiers/SizeModifiers.kt
+++ b/frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/ui/modifiers/SizeModifiers.kt
@@ -98,3 +98,16 @@ fun Modifier.maxHeight(size: CSSLengthOrPercentageNumericValue): Modifier = styl
 fun Modifier.maxHeight(maxHeight: MaxHeight): Modifier = styleModifier {
     maxHeight(maxHeight)
 }
+
+fun Modifier.wrapContentWidth(): Modifier = styleModifier {
+    width(Width.FitContent)
+}
+
+fun Modifier.wrapContentHeight(): Modifier = styleModifier {
+    height(Height.FitContent)
+}
+
+fun Modifier.wrapContentSize(): Modifier = styleModifier {
+    width(Width.FitContent)
+    height(Height.FitContent)
+}


bitspittle (commented):
So, I'm struggling a bit with this review, mostly because I don't have a deep understanding of what the `wrapContent*` methods do.

Note that the [wrapContentWidth](https://developer.android.com/reference/kotlin/androidx/compose/foundation/layout/package-summary#(androidx.compose.ui.Modifier).wrapContentWidth(androidx.compose.ui.Alignment.Horizontal,kotlin.Boolean)) etc. methods have more parameters. I wonder how much value it is adding these methods if we don't support the alignment and bounding parameters ourselves. (And I'm not sure it's possible / makes sense to support those parameters in a Compose HTML world)

I'm worried that the Jetpack Compose methods here are fundamentally different than the CSS approach we're replacing them with. If we get this wrong, then we'll be giving users the wrong mental model, and worst case people who find the behavior different from what they expected will file bugs, at which point we might not be able to change things easily due to legacy code.

How did you come up with these implementations? There's a good chance you understand the Jetpack Compose methods and the FitContent web concepts more deeply than I do.

ShreckYe:
I am actually not very sure about whether the Compose `wrapContent...` behaves the same as [the CSS `fit-content`](https://developer.mozilla.org/en-US/docs/Web/CSS/fit-content) either. I have been developing [a project](https://github.com/huanshankeji/compose-multiplatform-material) that delegates to both `androidx.compose` and Compose HTML that also depends on your work, in which as I tested they behave similarly in the way that they shrink a layout to its contents, but I do think that there are some differences.

From the [wrapContentWidth docs](https://developer.android.com/reference/kotlin/androidx/compose/foundation/layout/package-summary#(androidx.compose.ui.Modifier).wrapContentWidth(androidx.compose.ui.Alignment.Horizontal,kotlin.Boolean)):

> Allow the content to measure at its desired width without regard for the incoming measurement [minimum width constraint](https://developer.android.com/reference/kotlin/androidx/compose/ui/unit/Constraints#minWidth()), and, if [unbounded](https://developer.android.com/reference/kotlin/androidx/compose/foundation/layout/package-summary#(androidx.compose.ui.Modifier).wrapContentWidth(androidx.compose.ui.Alignment.Horizontal,kotlin.Boolean)) is true, also without regard for the incoming measurement [maximum width constraint](https://developer.android.com/reference/kotlin/androidx/compose/ui/unit/Constraints#maxWidth()). If the content's measured size is smaller than the minimum width constraint, [align](https://developer.android.com/reference/kotlin/androidx/compose/foundation/layout/package-summary#(androidx.compose.ui.Modifier).wrapContentWidth(androidx.compose.ui.Alignment.Horizontal,kotlin.Boolean)) it within that minimum width space. If the content's measured size is larger than the maximum width constraint (only possible when [unbounded](https://developer.android.com/reference/kotlin/androidx/compose/foundation/layout/package-summary#(androidx.compose.ui.Modifier).wrapContentWidth(androidx.compose.ui.Alignment.Horizontal,kotlin.Boolean)) is true), [align](https://developer.android.com/reference/kotlin/androidx/compose/foundation/layout/package-summary#(androidx.compose.ui.Modifier).wrapContentWidth(androidx.compose.ui.Alignment.Horizontal,kotlin.Boolean)) over the maximum width space.

From [this Stack Overflow answer](https://stackoverflow.com/a/45972381/5082913):

> **fit-content uses max-content**, *unless* `available < max-content`, then it uses **available**. *Unless* `available < min-content`, then it uses **min-content**.

So as I see from the docs they are quite different. It seems to me that Jetpack Compose follows [a top-down approach](https://developer.android.com/develop/ui/compose/layouts/constraints-modifiers#constraints-passed) where it passes size constraints from top to bottom, and HTML rendering by default follows a bottom-up approach where children/content sizes are determined before those of parents/containers. I think this is due to that Compose inherently renders on a bounded screen, while HTML inherently renders on unbounded space enabled by scrollbars. And as stated in the docs modifiers act as chained wrappers that wrap the component, while CSS styles act as properties of the component itself. In the case of `wrapContentWidth` and `fit-content`, when there isn't a passed minimum width constraint from the parent on Compose as in most cases, `wrapContentSize` will make the containing layout the size of `min(content, constraint_max)`, which is just similar to `min(max-content, available)` for `fit-content`; but in the case when the content is bigger than the containing layout and especially when `unbouned` is true, Compose only shows part of the content (by default at the center), but the browser expands the containing layout and thus the whole webpage, adding scrollbars when the body exceeds the browser window. As it seems to me the Compose behavior is similar to setting [CSS overflow](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow).

However, there are also similar inconsistencies with `Box`, `Row`, `Column`, and `fillMaxWidth/Height/Size` which are not that obvious. As I tested in my project the `Box` composable (and `Row` and `Column`) seems to shrink to its content size on Compose but the Kobweb `Box` doesn't, and the `fillMaxWidth/Height/Size` modifier inflates the containing layout to the constraint max while the Kobweb one doesn't. And as Compose and HTML are rendered differently internally there may not be a perfect answer. This may also be related to [block and inline HTML elements](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_flow_layout/Block_and_inline_layout_in_normal_flow).

> I'm worried that the Jetpack Compose methods here are fundamentally different than the CSS approach we're replacing them with. If we get this wrong, then we'll be giving users the wrong mental model, and worst case people who find the behavior different from what they expected will file bugs, at which point we might not be able to change things easily due to legacy code.

So I do agree with you on this and I am no expert on browser engines either. Maybe we should wait for someone who's more specialized in both Compose and browser rendering to comment on this.

ShreckYe:
This question [android - Match Width of Parent in Column (Jetpack Compose) - Stack Overflow](https://stackoverflow.com/questions/65942711/match-width-of-parent-in-column-jetpack-compose) might also be related.

bitspittle:
Thanks for the extra information! It's been very interesting to read.

It's true that the `Box`, `Column`, and `Row` concepts do deviate in some ways from some specific Android use-cases, but it's definitely a VERY useful concept to have in Kobweb anyway, because in a lot of cases, it does work close enough to what Android devs are expecting, *and* it's way way way easier to use them than flex layouts. 

In contrast, `wrapContentWidth` is I believe a kind of niche concept (let me know if I'm wrong!) and `width(Width.FitContent)`, our current replacement for it, actually reads pretty well to me.

I think I'd be more inclined to add stuff like `wrapContentWidth` (and other Jetpack Compose concepts that are currently missing) if I'm convinced people use it a lot in Android code *and especially* if the equivalent HTML code we would replace it with is complex.

Of course the calculations can change here if we find, as you said, someone else chiming in with a more nuanced implementation for these methods OR if I end up learning that `wrapContentWidth` is used just everywhere across all the Android codebases :)

PR #514, opened 2024-04-23
Migrate `<head>` configuration to be serializable

DennisTsar:


bitspittle (approved):

tools/gradle-plugins/core/src/main/kotlin/com/varabyte/kobweb/gradle/core/util/HtmlUtil.kt
```diff
@@ -0,0 +1,19 @@
+package com.varabyte.kobweb.gradle.core.util
+
+import kotlinx.html.HEAD
+import kotlinx.html.TagConsumer
+import kotlinx.html.stream.createHTML
+
+object HtmlUtil {
+    // Generate the html nodes without the containing <head> tag
```
bitspittle, line 8: Might be nice to say why we're doing this, without requiring people to leave the IDE and visit a GitHub issue. (Essentially, our intention is to accumulate all elements and wrap it with a <head> tag ourselves later)

bitspittle (approved):

PR #517, opened 2024-04-28
Update build.yml to use Gradle Caching and build timeout

ellet0:
See this [link](https://github.com/actions/setup-java?tab=readme-ov-file#caching-gradle-dependencies)
the timeout in case there is a infinite loop or some build bug (or if it's stuck for some reason) so it won't work forever, there is a limit (it's higher in public/open source projects)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,7 @@ env:
 jobs:
   build:
     runs-on: ubuntu-latest
+    timeout-minutes: 5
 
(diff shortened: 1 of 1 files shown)


DennisTsar:
The `setup-gradle` action does its own caching and recommends to not use `cache: gradle`: https://github.com/gradle/actions/blob/main/docs/setup-gradle.md#incompatibility-with-other-caching-mechanisms

PR #521, opened 2024-05-04
Add `CssStyle` tests

DennisTsar:


frontend/silk-foundation/src/jsTest/kotlin/com/varabyte/kobweb/silk/CssStyleTest.kt
```diff
@@ -0,0 +1,137 @@
...
+    fun componentVariantStylesSheetOrder() {
+        // Styles must be non-empty to be registered
+        val stylesheet = StyleSheet()
+        val BaseStyle = CssStyle.base<ComponentKind> { Modifier.color(Colors.Red) }
+        val BaseStyleName = "base-style"
+        val Variant = BaseStyle.addVariantBase { Modifier.color(Colors.Red) }
+        val VariantName = "secondary-style"
+        _SilkTheme = MutableSilkTheme().apply {
+            // Intentionally register out of order
+            registerVariant(VariantName, Variant)
```
bitspittle, line 38: Optional: You can just do `registerVariant("secondary-style", Variant)` and then use `SilkTheme.nameFor(Variant)` as a way to avoid these capitalized variables, reduce a few lines, and also test that `SilkTheme.nameFor(...)` is WAI.

In this case, I would have one test verify that the names are the same as what went into the registerVariant / registerStyle methods, but then after that, the remaining tests can just assume it works.

bitspittle (approved):
Very cool to finally have a way to do what I thought was impossible.

At some point we might want to use the ComposeUtil class to a new module, maybe some compose-test-support test only module or something. But it doesn't have to be now!

frontend/silk-foundation/src/jsTest/kotlin/com/varabyte/kobweb/silk/CssStyleTest.kt
DennisTsar: Good call, updated.

bitspittle (approved):

PR #531, opened 2024-05-13
Create update-gradle-wrapper.yml to automate the process of updating Gradle wrapper

ellet0:
This PR doesn't do anything, it use all the code from [Update Gradle Wrapper Action](https://github.com/marketplace/actions/update-gradle-wrapper-action) from the GitHub marketplace, we often have too many versions for different tools and dependencies, this would make the process of keep tracking of all of them and ensuring them they are up to date to benefit from latest bug fixes, improvements, features and security fixes is a more time consuming and sometimes we forgot about them

If you read this [link](https://github.com/marketplace/actions/update-gradle-wrapper-action#hygiene):

>Gradle is under heavy development, with new releases available every few weeks. Projects that stick to old Wrapper versions can't > benefit from the tons of features and bug fixes being rolled out.
> 
> Updating the build system only once or twice per year might become a tedious task. You will need to walk through longer changelogs > and handle any breaking change. Updating frequently helps do less work, take advantage of new features earlier and safely drop > deprecated functionality.
>
> This action runs automatically at the declared schedule and creates Pull Requests with detailed information about the version update.

It won't automatically update the Gradle wrapper version, it will send a PR and have to manually accept or reject the change, and it will be scheduled

### DIFF
--- a/.github/workflows/update-gradle-wrapper.yml
+++ b/.github/workflows/update-gradle-wrapper.yml
@@ -1,0 +1,15 @@
+name: Update Gradle Wrapper
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+  update-gradle-wrapper:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Update Gradle Wrapper
+        uses: gradle-update/update-gradle-wrapper-action@v1


bitspittle:
I appreciate the intention behind this suggestion, but it strikes me as something that will generate noise. Both in pinging us with PRs when we're otherwise busy, but also generating a bunch of extra jobs in our CI pipeline history.

Updating the gradle wrapper version is so trivial to do, and we've been doing it on our own terms, with no issues so far. I'm really hesitant to take a change that doesn't solve a problem that we've actually had.

ellet0:
> I appreciate the intention behind this suggestion, but it strikes me as something that will generate noise. Both in pinging us with PRs when we're otherwise busy, but also generating a bunch of extra jobs in our CI pipeline history.
> 
> Updating the gradle wrapper version is so trivial to do, and we've been doing it on our own terms, with no issues so far. I'm really hesitant to take a change that doesn't solve a problem that we've actually had.

It doesn't necessarily fix an issue, but for me, it's useful to have, especially when I have too many projects 

But what would be more useful is to use a lint tool just like eslint in JS ecosystem, I heard of [KtLint](https://github.com/pinterest/ktlint) but only start using it recently, it use all the rules and guidelines from official Kotlin but even more strict and can catch some quality and issues that the compiler can't 

It also has a plugin for Intellij and can be used in CI (just by using the build task)

It has a Gradle plugin, and you only have to fix the issues and the warrings it gives you the first time, and then it should give you the warrning in the IDE

PR #536, opened 2024-05-17
Add support for multiple box shadows on BoxShadow Modifier API

rafaeltonholo:
Currently, the BoxShadow Modifier API only supports a single shadow. This PR enables the API to receive multiple shadows or `none`, in case the user wants to remove a shadow in a variant style.

### DIFF
--- a/frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Box.kt
+++ b/frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Box.kt
@@ -40,11 +40,12 @@ class BoxSizing private constructor(private val value: String) : StylePropertyVa
     }
 }
 
+// See: https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing
 fun StyleScope.boxSizing(boxSizing: BoxSizing) {
     boxSizing(boxSizing.toString())
 }
 
-// See: https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing
+// See: https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow
 fun StyleScope.boxShadow(value: String) {
     property("box-shadow", value)
 }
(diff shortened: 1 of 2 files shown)


frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Box.kt
```diff
@@ -57,7 +58,61 @@ fun StyleScope.boxShadow(
...
+
+fun StyleScope.boxShadow(vararg shadows: CSSBoxShadow) {
+    boxShadow(shadows.toList())
+}
+
+fun StyleScope.boxShadow(shadows: List<CSSBoxShadow>) {
+    boxShadow(value = shadows.joinToString(transform = CSSBoxShadow::toString))
+}
+
+/**
```
bitspittle, line 81: Nit: Feels strange to me that this header comment doesn't begin with a summary line. How about something like:

```
/**
 * Data values relevant to the `box-shadow` CSS property.
 *
 * @property ...
 *
 * @see <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow">box-shadow documentation</a>
 */
```

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Box.kt
```diff
@@ -57,7 +58,61 @@ fun StyleScope.boxShadow(
...
+            inset = inset,
+        ),
+    )
+}
+
+fun StyleScope.boxShadow(vararg shadows: CSSBoxShadow) {
+    boxShadow(shadows.toList())
+}
+
+fun StyleScope.boxShadow(shadows: List<CSSBoxShadow>) {
```
bitspittle, line 77: I was on the fence about this, but for now, let's just leave the `List` version of these methods out. If we do it here, we'd have to update a bunch of other places to be consistent, and I'm not sure we need to commit to supporting two methods every time we want to add a CSS property that takes multiple arguments.

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Box.kt
```diff
@@ -57,7 +58,61 @@ fun StyleScope.boxShadow(
...
+    boxShadow(value = shadows.joinToString(transform = CSSBoxShadow::toString))
```
bitspittle, line 78: There should be code protecting against no items passed in.

```
if (shadows.isNotEmpty()) {
   boxShadow(...)
}
```

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Box.kt
```diff
@@ -40,11 +40,12 @@ class BoxSizing private constructor(private val value: String) : StylePropertyVa
     }
 }
 
+// See: https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing
 fun StyleScope.boxSizing(boxSizing: BoxSizing) {
     boxSizing(boxSizing.toString())
 }
 
-// See: https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing
+// See: https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow
```
bitspittle, line 48: No action required: Thank you for fixing this!

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Box.kt
```diff
@@ -57,7 +58,61 @@ fun StyleScope.boxShadow(
...
+data class CSSBoxShadow(
+    val offsetX: CSSLengthNumericValue = 0.px,
+    val offsetY: CSSLengthNumericValue = 0.px,
+    val blurRadius: CSSLengthNumericValue? = null,
+    val spreadRadius: CSSLengthNumericValue? = null,
+    val color: CSSColorValue? = null,
+    val inset: Boolean = false,
+) : CSSStyleValue {
+    companion object {
+        val None = "none".unsafeCast<CSSBoxShadow>()
```
bitspittle, line 112: Ah, this is subtle, but we shouldn't put these companion object values inside these parameter classes.

Check out Transition.kt and look at the `TransitionProperty` class as an example.

We will need to introduce a new `BoxShadow` class if we want to support `None` in this PR. If you do that, please also support `Inherit`, `Initial`, `Revert`, and `Unset`. Finally, you would add a `boxShadow` method that takes a single (non-vararg) `boxShadow: BoxShadow` parameter. 

(Note at this point you can't set "None" multiple times, or combine "None" with other box-shadow values, which is the more correct behavior. In other words, `box-shadow: 10px 5px 5px red, 60px -16px teal;` is valid but `box-shadow: 10px 5px 5px red, none, 60px -16px teal;` is not!)

You are also welcome to rip the companion object out for now, if you would rather I did it. 

For more context: The `CSS-` prefixed classes here represent code-only concepts that wrap things that are otherwise unnamed in the JavaScript standard library. They're meant to supplement the standard library in other words (with Kotlin type-safe functionality) rather than represent parts of it. 

Non "CSS-" prefixed names are what users are more likely to type intuitively. For example, if a user sees the docs for "box-shadow: none" then the expectation is they will type "BoxShadow.None" and it should autocomplete.

playground/site/src/jsMain/kotlin/playground/pages/Modifiers.kt
bitspittle, line 1: These examples are really nice and I do appreciate it. After thinking on it, I don't think we should be adding this to the playground, but I will use these soon to create a unit test class for your box-shadow APIs.

So let's just revert it here? I will give you credit when I add the test of course.

bitspittle (commented):
Thanks for this PR!

Phew, as you may have seen, 0.18.0 is finally out. Took quite a bit longer than I thought, so I wanted to say sorry for the delay.

Overall, this looks great, and I appreciate the time you spent on the header docs. I think we can get this into 0.18.1 fairly soon. There are some fixes that will have to happen first, but they should be fairly trivial, and hopefully helpful in understanding how nuanced this stuff can get sometimes.

Since this is our first time working together, just wanted to say feel free to ask questions or push back with your own suggestions anytime, happy to keep a dialog open.

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Box.kt
rafaeltonholo: I like the idea of protection against missing parameters. However, I suggest using `check(condition)` instead, as it would provide better reasoning for the user when consuming this API in the form of an error if you agree so.

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Box.kt
rafaeltonholo: I'm not entirely sure if I understand your point here. For example, when we look at `Transition.kt`, I can see why it makes sense to have a separate class for it in that situation.

The `transition` property in CSS is a shorthand property for 5 other properties. Because of this, it makes sense to me to have them separated from the `CSSTransition`, which represents that shorthand property.

On the other hand, `box-shadow` is a "complete" property, like `position`, `display`, or `color`, that can receive `none` or `<shadows>` as a value. So, in my understanding, it would make sense to have a `CSS-` definition for it, since there is no representation for that in the Kotlin/JS standard library. However, I'm not entirely sure if I really understand the context here.

When you mention having another class for `BoxShadow`, do you mean having both `CSSBoxShadow` and another one called `BoxShadowProperty`?

Something like:
```kotlin
value class BoxShadow private constructor(private val value: String) : StylePropertyValue {
    override fun toString(): String = value

    companion object {
        // Global values
        val None = BoxShadow("none")
        val Inherit = BoxShadow("inherit")
        val Initial = BoxShadow("initial")
        val Revert = BoxShadow("revert")
        val Unset = BoxShadow("unset")

        // Custom
        operator fun invoke(
            offsetX: CSSLengthNumericValue = 0.px,
            offsetY: CSSLengthNumericValue = 0.px,
            blurRadius: CSSLengthNumericValue? = null,
            spreadRadius: CSSLengthNumericValue? = null,
            color: CSSColorValue? = null,
            inset: Boolean = false,
        ): BoxShadow = BoxShadow(
            CSSBoxShadow(
                offsetX = offsetX,
                offsetY = offsetY,
                blurRadius = blurRadius,
                spreadRadius = spreadRadius,
                color = color,
                inset = inset
            ).toString(),
        )
    }
}

data class CSSBoxShadow(
    val offsetX: CSSLengthNumericValue = 0.px,
    val offsetY: CSSLengthNumericValue = 0.px,
    val blurRadius: CSSLengthNumericValue? = null,
    val spreadRadius: CSSLengthNumericValue? = null,
    val color: CSSColorValue? = null,
    val inset: Boolean = false,
) : CSSStyleValue { ... }
```

In that case, the API should receive `CSSBoxShadow` or `BoxShadow`?

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Box.kt
rafaeltonholo: I see, it makes sense to me know. Going to only add the if in that case

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Box.kt
bitspittle: Ah you're right! `transition` and `animation` are shorthand properties and `box-shadow` is just a regular one.

In that case, we might not need `CSSBoxShadow`. Instead, we can take a page from one of the more complex CSS properties we define that use sealed classes.

For two examples with vararg and non-vararg cases, look at `Content` and `FontVariantAlternates`. Following those ideas, we could provide an API like:

```kotlin
sealed class BoxShadow(private val value: String) : StylePropertyValue {
   private class Single(value: String) : BoxShadow(value)
   class Listable(value: String) : BoxShadow(value)

   companion object {
      fun of(
         val offsetX: ...,
         val offsetY: ...,
       ): Listable { /* ... toString logic of CSSBoxShadow here ... */ }

      // Keyword
      val None: BoxShadow = Single("None")
      
      // Global
      val Inherit = ...
     ...
   }
}

fun StyleScope.boxShadow(boxShadow: BoxShadow) { ... }
fun StyleScope.boxShadow(vararg boxShadows: BoxShadow.Listable) { ... }
```

and then callers would use:

```kotlin
Modifier.boxShadow(BoxShadow.of(...), BoxShadow.of(...))
```

Sorry for suggesting the wrong source examples in this case, I totally forgot about the short-hand vs. regular property situation here.

Does this suggestion make sense?

rafaeltonholo: Yes, that makes more sense now. Let me do some changes and I'll push it again :)

frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/ui/modifiers/BoxModifiers.kt
```diff
@@ -9,17 +9,175 @@ fun Modifier.boxDecorationBreak(boxDecorationBreak: BoxDecorationBreak) = styleM
...
+ * ```css
+ * @layer component-variant {
+ *     .silk-button-box-shadow {
+ *         box-shadow: box-shadow: rgba(0, 0, 0, 0.15) 0px 1px 3px 1px,
+ *                     rgba(0, 0, 0, 0.298) 0px 1px 2px 0px;
+ *     }
+ * }
+ * ```
+ */
+fun Modifier.boxShadow(vararg shadows: BoxShadow.Shadow): Modifier = styleModifier {
```
bitspittle, line 111: Not 100% sure what I feel about the name `BoxShadow.Shadow` but I certainly won't block the PR over it! We can easily tweak this name later if we come up with something else.

bitspittle (approved):
Looks great, and thanks for your patience on the back and forth.

I can tell by your attention to detail that wherever you work or whatever you work on is lucky to have you.

You may be interested to hear that the discussions we had here about the "CSS-" prefix has actually caused us to reflect on the pattern more broady in our codebase, and it seems likely that the way I designed CSSTransition and CSSAnimation is wrong.

> [!NOTE]
> CSSAnimation comes from Compose HTML but they screwed up their implementation, which is originally what caused me to introduce a fixed version of it into Kobweb, but now we're thinking `Animation.of` is more Kobweb idiomatic.

frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/ui/modifiers/BoxModifiers.kt
rafaeltonholo: I've named it `Shadow` because that's how [W3C](https://www.w3.org/TR/css-backgrounds-3/#box-shadow) refers to that property value. However, MDN uses `spread-shadow` instead, which could also be an alternative. Since W3C defines the web standard, I prefer sticking with their name. But I understand that it looks quite ugly as `BoxShadow.Shadow`.

Now that I think about it, maybe I should name it `BoxShadow.Property` instead, which is similar to what you're already doing with the other properties in the project. If you'd like, I can create a small PR to address that, or we can leave it as is and address it later.

rafaeltonholo:
> Looks great, and thanks for your patience on the back and forth.
> 
> I can tell by your attention to detail that wherever you work or whatever you work on is lucky to have you.
> 
> You may be interested to hear that the discussions we had here about the "CSS-" prefix has actually caused us to reflect on the pattern more broady in our codebase, and it seems likely that the way I designed CSSTransition and CSSAnimation is wrong.
> 
> Note
> 
> CSSAnimation comes from Compose HTML but they screwed up their implementation, which is originally what caused me to introduce a fixed version of it into Kobweb, but now we're thinking `Animation.of` is more Kobweb idiomatic.

I'm glad that I could help with the project! I really like its idea and I wish I can contribute more in the future :)

PR #538, opened 2024-05-18
Improve accesibility via adding aria-label to the switch input

carlosgub:
This pull request addresses an accessibility issue identified using the Lighthouse tool. The problem occurs due to the Switch Component lacking a proper label that describes the input used as a switch.

To resolve this, I added the modifier `arialLabel` to the Input Component. This solution I get it from here [aria-label accesibility](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label)

You can see similar problems discussed in the following links:

[Deque University: Accessibility Rule - Ensure every form element has a label](https://dequeuniversity.com/rules/axe/4.8/label)

### DIFF
--- a/README.md
+++ b/README.md
@@ -4877,6 +4877,8 @@ limitations:
   the browser for security reasons and can only be set through HTML / CSS).
 * you won't have the ability to turn elements on / off when printing the page.
 * accessibility tools for browsers might not work.
+* the download size of the rendering components is not insignificant and apparently not very compressible, often
+  resulting in a site's basic footprint being 4-6x larger total (e.g. 200-400K vs. 2-3MB for small sites).
 
 It would also prevent a developer from making use of the rich ecosystem of Javascript libraries out there.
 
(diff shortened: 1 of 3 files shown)


bitspittle:
Hey Carlos,

Sorry for pinging this PR like a year after you put it up. This has been a very bad job on my part.

I'm getting old, but I thought I had left comments on this back in the day. Maybe we talked elsewhere, like in the Kotlin slack or something?

Anyway, I feel bad we have kept this PR open without ever applying it; that said, I realize shouldn't keep kicking this can down the road. So I've filed https://github.com/varabyte/kobweb/issues/673 (blocking the 1.0 milestone) and I'm going to close this PR for now. (Feel free to leave comments if you felt that this was a bad decision)

My thinking is this issue is something we have to figure out more holistically across all widgets, and I'm not sure I want to do the font-size 0 trick. Feel free to subscribe to the bug so you can get updates for when we finally have a chance to work on it.

Thanks for bringing this issue to our attention!

PR #541, opened 2024-05-30
Add support for `Arrangement.spacedBy`

rafaeltonholo:
## Description
The current implementation of the `Arrangement` API is missing the `spacedBy` function, provided by Compose UI. This PR enables the API to support `Arrangement` with spacing between the children elements for the given container.

> [!IMPORTANT]
> Compose UI also supports negative values for the `spacedBy` function. As we are using `gap` for adding the space between the children of a `Colunm` or a `Row`, and `gap` only accepts values within [0, Infinity], we are ignoring negative values.

## Comparison with Jetpack Compose
The following examples compare the implementation on Kobweb and Jetpack Compose on Android. The structure of the Composable will be:
```
Column(600x600, border 1 solid black)
    Row(fillMaxWidth, verticalAlignment = Alignment.CenterVertically) x 4
        Button
```

<table style="width: 100%">
  <thead>
    <tr>
      <th>Kobweb</th>
      <th>Jetpack Compose</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td colspan="2">
        <pre><code>Column(
verticalArrangement = 
  Arrangement.spacedBy(space = 16.px)
)
  Row(/*no-gap/spaced-by*/) x4</code></pre>
      </td>
    </tr>
    <tr>
      <td>
        
      </td>
      <td>
        
      </td>
    </tr>
    <tr>
      <td colspan="2">
        <pre><code>Column(
verticalArrangement = 
  Arrangement.spacedBy(
    space = 16.px, 
    alignment = Alignment.CenterVertically,
  )
)
  Row(/*no-gap/spaced-by*/) x4</code></pre>
      </td>
    </tr>
    <tr>
      <td>
        
      </td>
      <td>
        
      </td>
    </tr>
    <tr>
      <td colspan="2">
        <pre><code>Column(
verticalArrangement = 
  Arrangement.spacedBy(
    space = 16.px, 
    alignment = Alignment.Bottom,
  )
)
  Row(/*no-gap/spaced-by*/) x4</code></pre>
      </td>
    </tr>
    <tr>
      <td>
        
      </td>
      <td>
        
      </td>
    </tr>
    <tr>
      <td colspan="2">
        <pre><code>Column(
verticalArrangement = 
  Arrangement.spacedBy(
    space = 16.px, 
    alignment = Alignment.CenterVertically,
  )
)
  Row(
    horizontalArrangement =
      Arrangement.spacedBy(space = 16.px),
  )
  Row(
    horizontalArrangement =
      Arrangement.spacedBy(
        space = 16.px,
        alignment = Alignment.Start,
      ),
  )
  Row(
    horizontalArrangement =
      Arrangement.spacedBy(
        space = 16.px,
        alignment = Alignment.CenterHorizontally,
      ),
  )
  Row(
    horizontalArrangement =
      Arrangement.spacedBy(
        space = 16.px,
        alignment = Alignment.End,
      ),
  )</code></pre>
      </td>
    </tr>
    <tr>
      <td>
        
      </td>
      <td>
        
      </td>
    </tr>
  </tbody>
</table>

### DIFF
--- a/frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/foundation/layout/Arrangement.kt
+++ b/frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/foundation/layout/Arrangement.kt
@@ -243,8 +243,8 @@ object Arrangement {
  * @property spacing The spacing value used in the arrangement.
  */
 internal sealed class SpacedAligned(
-    open val spacing: CSSSizeValue<out CSSUnitLength>,
-) {
+    override val spacing: CSSSizeValue<out CSSUnitLength>,
+) : Arrangement.HorizontalOrVertical {
     /**
      * The CSS classes applied to the arrangement.
      */
@@ -258,7 +258,7 @@ internal sealed class SpacedAligned(
  */
 internal data class SpacedAlignedHorizontalOrVertical(
     override val spacing: CSSSizeValue<out CSSUnitLength>,
-) : SpacedAligned(spacing), Arrangement.HorizontalOrVertical {
+) : SpacedAligned(spacing) {
     // Custom classes for default alignment applied for spacing and alignment
     override val classes = arrayOf(KOBWEB_ARRANGE_SPACED_BY, KOBWEB_ARRANGE_START)
 }
--- a/frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/foundation/layout/Arrangement.kt
+++ b/frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/foundation/layout/Arrangement.kt
@@ -248,7 +248,7 @@ internal sealed class SpacedAligned(
     /**
      * The CSS classes applied to the arrangement.
      */
-    abstract val classes: Array<String>
+    abstract val classNames: Array<String>
 }
 
 /**
@@ -260,7 +260,7 @@ internal data class SpacedAlignedHorizontalOrVertical(
     override val spacing: CSSSizeValue<out CSSUnitLength>,
 ) : SpacedAligned(spacing) {
     // Custom classes for default alignment applied for spacing and alignment
-    override val classes = arrayOf(KOBWEB_ARRANGE_SPACED_BY, KOBWEB_ARRANGE_START)
+    override val classNames = arrayOf(KOBWEB_ARRANGE_SPACED_BY, KOBWEB_ARRANGE_START)
 }
 
 /**
@@ -274,7 +274,7 @@ internal data class SpacedVerticalAligned(
     val alignment: Alignment.Vertical,
 ) : SpacedAligned(spacing), Arrangement.Vertical {
     // Custom classes for vertical alignment based on the alignment type
-    override val classes: Array<String> = arrayOf(
+    override val classNames: Array<String> = arrayOf(
         KOBWEB_ARRANGE_SPACED_BY,
         when (alignment) {
             Alignment.Bottom -> KOBWEB_ARRANGE_BOTTOM
@@ -296,7 +296,7 @@ internal data class SpacedHorizontalAligned(
     val alignment: Alignment.Horizontal,
 ) : SpacedAligned(spacing), Arrangement.Horizontal {
     // Custom classes for horizontal alignment based on the alignment type
-    override val classes: Array<String> = arrayOf(
+    override val classNames: Array<String> = arrayOf(
         KOBWEB_ARRANGE_SPACED_BY,
         when (alignment) {
             Alignment.Start -> KOBWEB_ARRANGE_START
--- a/frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/foundation/layout/Arrangement.kt
+++ b/frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/foundation/layout/Arrangement.kt
@@ -1,18 +1,74 @@
 package com.varabyte.kobweb.compose.foundation.layout
 
-object Arrangement {
-    sealed interface Horizontal
-    sealed interface Vertical
-    sealed interface HorizontalOrVertical : Horizontal, Vertical
+import com.varabyte.kobweb.compose.css.*
+import com.varabyte.kobweb.compose.style.KOBWEB_ARRANGE_BOTTOM
+import com.varabyte.kobweb.compose.style.KOBWEB_ARRANGE_CENTER
+import com.varabyte.kobweb.compose.style.KOBWEB_ARRANGE_END
+import com.varabyte.kobweb.compose.style.KOBWEB_ARRANGE_FROM_STYLE
+import com.varabyte.kobweb.compose.style.KOBWEB_ARRANGE_SPACED_BY
+import com.varabyte.kobweb.compose.style.KOBWEB_ARRANGE_START
+import com.varabyte.kobweb.compose.style.KOBWEB_ARRANGE_TOP
+import com.varabyte.kobweb.compose.ui.Alignment
+import org.jetbrains.compose.web.css.*
 
-    object End : Horizontal
-    object Start : Horizontal
-    object Top : Vertical
-    object Bottom : Vertical
-    object Center : HorizontalOrVertical
-    object SpaceEvenly : HorizontalOrVertical
-    object SpaceBetween : HorizontalOrVertical
-    object SpaceAround : HorizontalOrVertical
+/**
+ * Used to specify the arrangement of the layout's children in layouts like
+ * [Row] or [Column] in their main axis direction (horizontal and vertical,
+ * respectively).
+ */
+object Arrangement {
+    /**
+     * Specifies the horizontal arrangement of the children. Mostly used on [Row].
+     */
+    sealed interface Horizontal {
+        /**
+         * Spacing added between any two adjacent children.
+         * defaults to `0.px`
+         */
+        val spacing: CSSLengthNumericValue get() = 0.px
+    }
+
+    /**
+     * Specifies the vertical arrangement of the children. Mostly used on [Column].
+     */
+    sealed interface Vertical {
+        /**
+         * Spacing added between any two adjacent children.
+         * defaults to `0.px`
+         */
+        val spacing: CSSLengthNumericValue get() = 0.px
+    }
+
+    /**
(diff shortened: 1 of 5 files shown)


frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/foundation/layout/Arrangement.kt
```diff
@@ -1,18 +1,87 @@
...
+import com.varabyte.kobweb.compose.style.KOBWEB_ARRANGE_TOP
+import com.varabyte.kobweb.compose.ui.Alignment
+import org.jetbrains.compose.web.css.*
+
+/**
+ * Used to specify the arrangement of the layout's children in layouts like
+ * [Row] or [Column] in their main axis direction (horizontal and vertical,
+ * respectively).
+ */
+@Immutable
```
bitspittle, line 20: Were these `@Immutable` / `@Stable` annotations required?

My impression is that using these annotations is a code smell and can cause fragility for future maintenaners who accidentally add something in the future that breaks the contract.

See also: https://developer.android.com/develop/ui/compose/performance/stability/fix#annotate-stable

> If it is possible to make your class stable without an annotation, you should strive to achieve stability that way.

frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/foundation/layout/Arrangement.kt
```diff
@@ -35,5 +104,226 @@ object Arrangement {
...
+ * @property spacing The spacing value used in the arrangement.
+ */
+@Immutable
+internal sealed class SpacedAligned(
+    open val spacing: CSSSizeValue<out CSSUnitLength>,
+) {
+    /**
+     * The CSS classes applied to the arrangement.
+     */
+    abstract val classes: Array<String>
```
bitspittle, line 269: Let's call this "classNames" to be consistent with the rest of Kobweb.

Also `toClasses()` -> `toClassNames` and `rowClassNames` and `columnClassNames`

(If there's a reason you disagree, maybe this is what they use in Jetpack Compose, feel free to discuss your thoughts here. Another approach is we can change `Modifier.classNames` to `Modifier.classes`, but originally I tend to avoid using properties / methods named "class" because of its ambiguity with Kotlin  / Java classes and its reserved keyword, e.g. it's ugly to write `for clazz in classes`)

frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/style/Arrangement.kt
```diff
@@ -1,23 +1,50 @@
...
+internal const val KOBWEB_ARRANGE_END = "kobweb-arrange-end"
+internal const val KOBWEB_ARRANGE_FROM_STYLE = "kobweb-arrange-from-style"
+internal const val KOBWEB_ARRANGE_SPACED_BY = "kobweb-arrange-spaced-by"
+internal const val KOBWEB_ARRANGE_SPACE_AROUND = "kobweb-arrange-space-around"
+internal const val KOBWEB_ARRANGE_SPACE_BETWEEN = "kobweb-arrange-space-between"
+internal const val KOBWEB_ARRANGE_SPACE_EVENLY = "kobweb-arrange-space-evenly"
+internal const val KOBWEB_ARRANGE_START = "kobweb-arrange-start"
+internal const val KOBWEB_ARRANGE_TOP = "kobweb-arrange-top"
+
+internal val ArrangeSpacedByValue by StyleVariable<CSSNumeric>(prefix = "kobweb")
```
bitspittle, line 21: No action required: Great job using stylevariable here! It's definitely the right way forward here, and I appreciate that you found the right tool for the job. (I guess I assume very few people actually use or are even aware of style variables)

bitspittle (commented):
Love the overall commit, and very very impressed (again) with docs and even the docs used for the PR message.

My biggest question is about the need for @Immutable / @Stable, that worries me a little bit. Were these changes required or were you just trying to annotate things because you thought it was best practice / it's what the Compose codebase does?

frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/foundation/layout/Arrangement.kt
```diff
@@ -35,5 +104,226 @@ object Arrangement {
...
+        SpacedHorizontalAligned(space, alignment)
+}
+
+/**
+ * Sealed class representing a spaced an [Arrangement] with a given spacing.
+ *
+ * @property spacing The spacing value used in the arrangement.
+ */
+@Immutable
+internal sealed class SpacedAligned(
```
bitspittle, line 263: In Jetpack Compose, [SpacedAligned inherits from HorizontalOrVertical](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/foundation/foundation-layout/src/commonMain/kotlin/androidx/compose/foundation/layout/Arrangement.kt;l=564?q=file:androidx%2Fcompose%2Ffoundation%2Flayout%2FArrangement.kt%20class:androidx.compose.foundation.layout.Arrangement).

It seems like these changes work (thanks to @DennisTsar for exploring this), anything we're missing?

```kotlin
internal sealed class SpacedAligned(
    override val spacing: CSSSizeValue<out CSSUnitLength>,
) : Arrangement.HorizontalOrVertical { ... }

internal class SpacedAlignedHorizontalOrVertical(
    spacing: CSSSizeValue<out CSSUnitLength>,
) : SpacedAligned(spacing) { ... }

internal class SpacedVerticalAligned(
    override val spacing: CSSSizeValue<out CSSUnitLength>,
    val alignment: Alignment.Vertical,
) : SpacedAligned(spacing) { ... }

internal class SpacedHorizontalAligned(
    override val spacing: CSSSizeValue<out CSSUnitLength>,
    val alignment: Alignment.Horizontal,
) : SpacedAligned(spacing) { ... }
```

and remove the code here:

```diff
fun Arrangement.Horizontal.toClasses() = when (this) {
    Arrangement.End -> arrayOf(KOBWEB_ARRANGE_END)
    Arrangement.Start -> arrayOf(KOBWEB_ARRANGE_START)
    is Arrangement.HorizontalOrVertical -> this.toClasses()
-   is SpacedHorizontalAligned -> classes
}

fun Arrangement.Vertical.toClasses() = when (this) {
    Arrangement.Top -> arrayOf(KOBWEB_ARRANGE_TOP)
    Arrangement.Bottom -> arrayOf(KOBWEB_ARRANGE_BOTTOM)
    is Arrangement.HorizontalOrVertical -> toClasses()
-   is SpacedAligned -> classes
-   is SpacedVerticalAligned -> classes
}
```

frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/foundation/layout/Arrangement.kt
rafaeltonholo: TL;DR: If not marked as `@Stable`/`@Immutable`, the composable that uses it as a parameter will trigger unnecessary recompositions.

<details>
<summary>Long explanation</summary>

After this comment, I questioned myself about whether or not using those annotations here. So I delved into their code to better understand why they are using `@Immutable` and `@Stable` in the `Arrangement`. Here are some thoughts I have on this.

Even though objects are stable to Compose when using `val`, the `Arrangement` class uses `val` inside it to create anonymous objects, thus the Compose can't really say if it is stable. However, that wouldn't be the case on Kobweb since we are using `data object` instead of `val`s when declaring the "subtypes" of `Arrangement`, but I would like to confirm, so I run a few tests/analysis to make sure about it.

Just for the purpose of analysis, I took the same code from [Android's `Arrangement`](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/foundation/foundation-layout/src/commonMain/kotlin/androidx/compose/foundation/layout/Arrangement.kt), placed it into an app, and removed all `@Immutable` and `@Stable` annotations. This is the result reported by the compiler:
```
stable class Absolute {
  runtime val Left: Horizontal
  runtime val Center: Horizontal
  runtime val Right: Horizontal
  runtime val SpaceBetween: Horizontal
  runtime val SpaceEvenly: Horizontal
  runtime val SpaceAround: Horizontal
  <runtime stability> = 
}
stable class MyArrangement {
  runtime val Start: Horizontal
  runtime val End: Horizontal
  runtime val Top: Vertical
  runtime val Bottom: Vertical
  runtime val Center: HorizontalOrVertical
  runtime val SpaceEvenly: HorizontalOrVertical
  runtime val SpaceBetween: HorizontalOrVertical
  runtime val SpaceAround: HorizontalOrVertical
  <runtime stability> = 
}
```

When migrating from `val`s to `data object`, similar to how we do it in Kobweb, the behavior of the report changes slightly, resulting in both `MyArrangement` and `Absolute` being `Stable`. However, when using it within a class or a composable, Compose still leaves it to the runtime to determine whether it is stable or not, resulting in its runtime stability being `Uncertain`.

The issue with using `Uncertain` is that it could trigger recomposition in places where it might not be necessary, as it is determined by the runtime whether to skip the composable or not.

By adding the `@Stable` annotation to the classes, it ensures Compose of its stability and prevents unnecessary recompositions, however, I wasn't certain if this applies to Kobweb, so I needed to run a few tests to confirm.

After removing the `@Immutable` and `@Stable` annotations and running a few tests in Kobweb, I noticed unnecessary recompositions occurring in Composables that used the `Arrangement` and didn't have any changes to trigger a recomposition.

This makes me believe that we should still annotate it as `@Immutable` and `@Stable`.

</stable>

rafaeltonholo: I'm going to remove the `@Stable` and `@Immutable` markers from this PR to keep the addition of the `spaceBy` separate from the recomposition issue fix

frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/foundation/layout/Arrangement.kt
rafaeltonholo: You are right! I have applied the change.

Thanks for bringing that :)

bitspittle (approved):

PR #551, opened 2024-06-13
Add support for `overscroll-behavior`

DennisTsar:


bitspittle (approved):
Looking at it like this, I like the "SingleValue" / "RepeatableValue" naming.

PR #558, opened 2024-06-24
Minor fix comment in css-style.md

kjeller:
When migrating I found a comment that confused me slightly.


bitspittle (approved):
Oh whoops, thank you!

PR #562, opened 2024-06-27
Remove `KobwebModuleTask`

DennisTsar:
It's no longer necessary now that tasks are explicitly configured with their necessary inputs.


bitspittle:
Very nice! I can appreciate that we don't have to duplicate the js with roots logic

bitspittle (approved):

PR #566, opened 2024-06-30
Add support for `window.ai`

dead8309:
# Desciption

Chrome is adding `window.ai` for embedding the Gemini Nano Ai model in Chrome natively.This pr adds the kotlin/js external declerations  for `AiManager` so we can use it in kobweb apps. I've also added an example Ai Page in playground

**Overview** - https://developer.chrome.com/docs/ai/built-in

### Playground Demo

https://github.com/varabyte/kobweb/assets/68665948/543459c1-2c1f-47ea-b131-72e4605c793e


bitspittle:
Ah, sorry if I wasn't clear in the discord server. I don't think this change belongs in the official Kobweb codebase -- we only want to support widely supported, cross-browser features here. I don't think it would be a good precedent to support a chrome-only feature.

However, if you want to publish a library from your own github repository, I'll be happy to surface it somewhere near the top of the README!

bitspittle:
Another option would be to create a template demo project. I'd be happy to include that in the kobweb-templates repo actually.

dead8309:
> we only want to support widely supported, cross-browser features here.

That makes sense. Perhaps in the future, if the APIs become standardized and widely adopted across browsers, you might consider including them.

I'll open a pr in templates repository to include a demo

PR #571, opened 2024-07-07
Fixed a formatting issue in readme at the section animations

FromWau:
Fixed a formatting issue in readme at the section [animations](https://github.com/FromWau/kobweb#animations)

Before fix:
image

After fix:
image


bitspittle:
Thank you!!

Note that the way we work is a little custom ([more details in here](https://github.com/varabyte/kobweb/blob/main/CONTRIBUTING.md)), and this should be based on the 0.18.3 branch.

I'm going to try to set this for you, but I've seen GitHub get really confused about this in the past, so if you see a bunch of random noise, I apologize in advance. I'll get your change in one way or another.

bitspittle:
Ah good, that worked seamlessly this time. Your change is in. Thanks again!

PR #573, opened 2024-07-18
Declare all inputs for markdown tasks

DennisTsar:
We mark lambda inputs with `@Nested`, which ensures tasks are rerun when the lambda is changed.


bitspittle (approved):

bitspittle (approved):
:heart:

PR #575, opened 2024-07-20
Watch for changes to build scripts in continuous mode

DennisTsar:


tools/gradle-plugins/core/src/main/kotlin/com/varabyte/kobweb/gradle/core/util/ProjectExtensions.kt
```diff
@@ -112,3 +114,15 @@ fun Project.toUidString(): String {
 
     return hash.joinToString("") { "%02x".format(it) }.uppercase()
 }
+
+/**
+ * Return the Gradle build script from the given project layout.
```
bitspittle, line 119: Nit: all Gradle build scripts under the given project layout (?)

bitspittle (approved):

PR #608, opened 2024-09-05
Add a README section about PageContext.isExporting

### LINKED ISSUE #607: Call out PageContext.isExporting in the README
Story:
Yesterday I used `PageContext.isExporting` to present a specific face to the site exporter. I needed this because the page is protected by authentication and redirects itself to index if not locally authenticated (saving a 401 response roundtrip). This leads to all the exported pages being essentially the login page, which in turn leads to login page flicker on first draw when the route is accessed directly.

- First I looked for a Gradle configuration option to disable snapshotting and instead present a specific static page for all routes. No can do.
- Second I looked for a way to recognize when a page is being exported. First place I look is the README which doesn't say anything about this. After a while I "accidentally" discovered the `isExporting` property.

Obviously now I'll remember it forever, but invoking IDE autocompletion on `PageContext` was not the first thing that I thought of when conducting my search.

-------

It might be beneficial to add a section about `PageContext.isExporting` to the README so that it's easier to find.

- [x] README

okarmazin:
Closes #607

Added as a subsection to `Exporting and running`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2497,6 +2497,35 @@ Once exported, you can test your site by running it locally before uploading. Yo
 `kobweb run --env prod --layout static` and a full stack site with `kobweb run --env prod --layout fullstack` (or just
 `kobweb run --env prod`).
 
+#### PageContext.isExporting
+
+Sometimes, you have behavior that should run when an actual user is navigating your site, but you don't want it to run
+at export time. For example, maybe you offer logged-in users an authenticated experience, but you'll never have 
+a logged-in user at export time.
+
+You can determine if your page is being rendered as part of an export by checking the `PageContext.isExporting`property.
+This gives you the opportunity to manipulate the exported HTML or avoid side effects associated with page loading.
+
+```kotlin
+@Composable
+fun AuthenticatedLayout(content: @Composable () -> Unit) {
+    var loggedInUser by remember { mutableStateOf<User?>(null) }
+
+    val ctx = rememberPageContext()
+    if (!ctx.isExporting) {
+        LaunchedEffect(Unit) { 
+            loggedInUser = checkForLoggedInUser() // <- A slow, expensive method  
+        }
+    }
+  
+    if (loggedInUser == null) { 
+        LoggedOutScaffold { content() }
+    } else { 
+        LoggedInScaffold(user) { content() } 
+    }
+}
+```
+
 ### Dynamic routes and exporting
 
 Dynamic routes are skipped over by the export process. After all, it's not possible to know all the possible values that


README.md
```diff
@@ -2497,6 +2497,30 @@ Once exported, you can test your site by running it locally before uploading. Yo
 `kobweb run --env prod --layout static` and a full stack site with `kobweb run --env prod --layout fullstack` (or just
 `kobweb run --env prod`).
 
+#### Use PageContext.isExporting for export recognition in your page
```
bitspittle, line 2500: Optional: Just "PageContext.isExporting" as the title here is fine I think; the section is short and the contents quickly explain what it will do.

README.md
```diff
@@ -2497,6 +2497,30 @@ Once exported, you can test your site by running it locally before uploading. Yo
...
+
+You can determine if your page is being exported by checking `PageContext.isExporting`. If `true`, the page is being
```
bitspittle, line 2502: How about an intro and a slight simplification of your text (since I think the "If `true`" sentence is not necessary to say as it's intuitive to understand that part):

---

Sometimes, you have behavior that should run when an actual user is navigating your site, but you don't want it to run at export time. For example, maybe you offer logged-in users an authenticated experience, but you'll never have a logged in user at export time.

You can determine if your page is being rendering as part of an export by checking `PageContext.isExporting` property. This gives you the opportunity to manipulate the exported HTML or avoid side effects associated with page loading.

README.md
```diff
@@ -2497,6 +2497,30 @@ Once exported, you can test your site by running it locally before uploading. Yo
...
+rendered for the purpose of site export.    
+This gives you the opportunity to manipulate the exported HTML or avoid side effects associated with page loading.
+
+```kotlin
+@Page
+@Composable
+fun AuthenticatedPage() {
+    val ctx = rememberPageContext()
+    if (ctx.isExporting) {
+        // We are exporting, the exporter is never authenticated.
```
bitspittle, line 2512: Nit: I think we can delete this first sentence. It's clarified from the previous text now.

README.md
```diff
@@ -2497,6 +2497,30 @@ Once exported, you can test your site by running it locally before uploading. Yo
...
+This gives you the opportunity to manipulate the exported HTML or avoid side effects associated with page loading.
+
+```kotlin
+@Page
+@Composable
+fun AuthenticatedPage() {
+    val ctx = rememberPageContext()
+    if (ctx.isExporting) {
+        // We are exporting, the exporter is never authenticated. 
+        // Present a lightweight skeleton without calling any APIs
```
bitspittle, line 2513: Nit: Missing "." at the end of the sentence.

README.md
```diff
@@ -2497,6 +2497,30 @@ Once exported, you can test your site by running it locally before uploading. Yo
...
+#### Use PageContext.isExporting for export recognition in your page
+
+You can determine if your page is being exported by checking `PageContext.isExporting`. If `true`, the page is being 
+rendered for the purpose of site export.    
+This gives you the opportunity to manipulate the exported HTML or avoid side effects associated with page loading.
+
+```kotlin
+@Page
+@Composable
+fun AuthenticatedPage() {
```
bitspittle, line 2509: Optional: Just thinking out loud here. I would probably make this `fun AuthenticatedLayout()` and not a page, since we'd want to encourage users to put authentication logic in a shared location and not have to repeat it per page.

That would look like:

```kotlin
@Composable
fun AuthenticatedLayout(content: @Composable () -> Unit) {
   var loggedInUser by remember { mutableStateOf<User?>(null) }

   val ctx = rememberPageContext()
   if (ctx.isExporting) {
      LaunchedEffect(Unit) {
         loggedInUser = checkForLoggedInUser() // <- A slow, expensive method
      }
  }
  
  if (loggedInUser == null) { 
     LoggedOutScaffold { content() }
  } else {
     LoggedInScaffold(user) { content() }
  }
}
```

WDYT?

README.md
```diff
@@ -2497,6 +2497,30 @@ Once exported, you can test your site by running it locally before uploading. Yo
...
+    val ctx = rememberPageContext()
+    if (ctx.isExporting) {
+        // We are exporting, the exporter is never authenticated. 
+        // Present a lightweight skeleton without calling any APIs
+        // We would get redirected to logout page after getting 401 from the API anyway...
```
bitspittle, line 2514: Nit: "We would get redirected to a logout page after getting a 401 from our API call anyway"

bitspittle (commented):
Thank you for adding this and helping me fill out the gaps in our README!

Left some suggestions. Let me know what you think and ALWAYS feel free to disagree or offer alternate suggestions of your own.

README.md
okarmazin: I looked around how other H4 headings are worded and saw they tend to be more sentence-like, that's why I wrote it like this.

A lone `PageContext.isExporting` is good too, changed.

README.md
okarmazin: Presenting the sample as a layout is a good call - the sample should nudge people the right way.

okarmazin: The snippet itself - it's just a different use case than the one I originally put in, and I have no thoughts either way.

Mine was written with the mindset that "nobody must ever see `content` if they are unauthenticated". And `isExporting` in that case prevents Kobweb from snapshotting `index`, where the unauthenticated exporter gets immediately redirected when navigating to the protected route.

I'm happy to change this other snippet in. Though I must say the first one with the early return guard clause seems a bit cleaner since it doesn't pollute the snippet with "user", all things being equal. And my bike's shed will be green, thank you very much!

Changed.

bitspittle (approved):
Thanks again!

PR #610, opened 2024-09-13
Update README for adding JAVA_HOME to Path variable

inditilve:


bitspittle (commented):
Hey, I appreciate the PR.

That said, I'm actually of the opinion that it is not our responsibility to help people set up their PATH variable. I think that it outside the scope of what the Kobweb README  should be responsible for. There is a lot of nuance in how and where people set up their system paths, and I don't think we want to get into it. Our README is large enough as it is already.

Note that you absolutely can use IntelliJ IDEA and compile Java / Kotlin / Kobweb projects without ever setting up your terminal to access java itself. And some systems may install Java for you and set the path themselves (I believe on Mac this happens for example, my system's java is set to `usr/bin/java`). So it might be a problem to have people mess with that if they don't know what they are doing.

PR #614, opened 2024-10-11
Update ktor to 3.0.0

DennisTsar:


bitspittle (approved):
Thanks for being on top of this!

Your approach to the keepalive delay is elegant code. When I was thinking of how to solve the issue when you first mentioned it, I assumed you would have used some sort of accumulating elapsed time value, because for some reason I didn't realize that structured concurrency was the perfect choice to use here. Probably my own neural net is trained too much on old, imperative, ugly ugly code :) Thanks for helping me see a better way.

PR #615, opened 2024-10-11
Add kotlinx serialization extension methods for HttpFetcher

dead8309:


bitspittle:
The API looks great!

One thing I am now realizing is something is still missing: in addition to `window.http.*`, we also have `window.api.*`, which is basically like `window.http` but designed for use with [server API routes](https://github.com/varabyte/kobweb?tab=readme-ov-file#define-api-routes). (In fact, `ApiFetcher` was first but I added `HttpFetcher` later since it was more generally applicable)

**So here's the question...**

Do we create a single serialization module, where covers both `HttpFetcher` and `ApiFetcher` extensions can live together? Or do we have a separate module for each one?

I'm actually leaning to using one module for both, because I think in practice it's unlikely for `HttpFetcher` and its extension methods ever to get upstreamed to JetBrains, so instead we can just create a single module which is simple and convenient for Kobweb users (and we can always break it up later if we need to).

But if we do that, it demands a different name. Maybe just `kobwebx-core-serialization` or something like that? (Since `ApiFetcher` lives in `kobweb-core`)

The two choices that I'm seeing:

* One module, which encompasses both `HttpFetcher` and `ApiFetcher`, but it needs a different name at that point.
* Two modules, one for browser, and the other for core. And the core module has an `api` dependency on the browser module. We'd still need a new name for the `ApiFetcher` one.

**Any opinions?**

dead8309:
I agree, I'm also leaning towards going with a single module approach. It will simplify things for Kobweb users and as you mentioned, we can always split it later if we need to.

For the internal structure, we can organize the HttpFetcher and ApiFetcher related code into separate packages within the module.

Do you have any thoughts on how we should proceed with implementing this? Should I start by outlining the internal structure of the new module?

dead8309:
Sorry! Skipped the [Latest branch](https://github.com/varabyte/kobweb/blob/main/CONTRIBUTING.md#latest-branch) part in Contributing file

bitspittle:
> For the internal structure, we can organize the HttpFetcher and ApiFetcher related code into separate packages within the module.
> 
> Do you have any thoughts on how we should proceed with implementing this? Should I start by outlining the internal structure of the new module?

I'll take a look at your commits in just a sec, but what you outlined in the first sentence -- one module with multiple packages in it -- sounds good to me.

> As for the name of the module maybe one of `kobwebx-core-serialization`, `kobwebx-serialization-extensions`. 

I'm currently leaning towards `kobwebx-core-serialization-kotlinx` (as there are other serialization libraries out there besides the kotlinx one). But that is a mouthful, so I will think a bit more about it. Feel free to let me know if you think it is awful :)

Still, you don't need to feel blocked by a name. We can always check it in with one thing and we still have time to come up with another name before releasing 0.19.3.

(Not something to worry about in this commit, but I will also be considering if maybe we should merge [the Kobweb worker serialization extension utilities](https://github.com/varabyte/kobweb/blob/main/common/kobwebx-serialization-kotlinx/src/commonMain/kotlin/com/varabyte/kobwebx/worker/kotlinx/serialization/util/KotlinSerializationUtils.kt) into whatever we're creating here)

playground/site/build.gradle.kts
```diff
@@ -36,12 +37,14 @@ kotlin {
...
         jsMain.dependencies {
             implementation(libs.compose.html.core)
+            implementation(libs.kotlinx.serialization.json)
             implementation("com.varabyte.kobweb:kobweb-core")
             implementation("com.varabyte.kobweb:kobweb-silk")
             implementation("com.varabyte.kobwebx:silk-icons-fa")
             implementation("com.varabyte.kobwebx:kobwebx-markdown")
             implementation(project(":sitelib"))
             implementation(project(":worker"))
+            implementation("com.varabyte.kobweb:kobwebx-browser-serialization")
```
bitspittle, line 47: Nit: Needs to be updated to latest name

frontend/kobweb-core/src/jsMain/kotlin/com/varabyte/kobweb/browser/ApiFetcher.kt
```diff
@@ -27,7 +27,7 @@ class ApiFetcher(private val window: Window) {
      */
     var logOnError: Boolean by window.http::logOnError
 
-    private fun toResource(apiPath: String, autoPrefix: Boolean): String {
+    fun toResource(apiPath: String, autoPrefix: Boolean): String {
```
bitspittle, line 30: Instead of exposing this method, let's keep it private and use the following pattern in the extension code:

```kotlin
suspend inline fun <reified T> ApiFetcher.get(
    apiPath: String,
    headers: Map<String, Any>? = null,
    abortController: AbortController? = null,
    autoPrefix: Boolean = true,
    serializer: DeserializationStrategy<T> = serializer()
): T = Json.decodeFromString(serializer, window.api.get(apiPath, headers, abortController, autoPrefix).decodeToString())
```

frontend/kobwebx-core-serialization/build.gradle.kts
```diff
@@ -0,0 +1,35 @@
...
+        jsTest.dependencies {
+            implementation(kotlin("test-js"))
+            implementation(libs.truthish)
+        }
+    }
+}
+
+kobwebPublication {
+    artifactId.set("kobwebx-core-serialization")
+    description.set("Generally useful Kotlinx Serialization extensions for the browser (not Node) APIs that could potentially move upstream someday.")
```
bitspittle, line 33: Let's leave out the upstream part of the comment, as that is no longer new.

Maybe just

`"Generally useful Kotlinx Serialization extensions for various Kobweb APIs."`

?

frontend/kobwebx-core-serialization/src/jsMain/kotlin/com/varabyte/kobweb/browser/ApiFetcherSerializationExt.kt
```diff
@@ -0,0 +1,261 @@
...
+import com.varabyte.kobweb.browser.http.tryPatch
+import com.varabyte.kobweb.browser.http.tryPost
+import com.varabyte.kobweb.browser.http.tryPut
+import kotlinx.browser.window
+import kotlinx.serialization.DeserializationStrategy
+import kotlinx.serialization.serializer
+
+
+/**
+ * Call GET on a target API path with [T] as the expected return type.
```
bitspittle, line 25: Optional: Perhaps add a bit about Kotlinx serialization, here and below?

```kotlin
/**
 * Call GET on a target API path with [T] (deserializalbe by Kotlinx serialization) as the expected return type.
 * ...
 */
```

frontend/kobwebx-core-serialization/src/jsMain/kotlin/com/varabyte/kobweb/browser/ApiFetcherSerializationExt.kt
bitspittle, line 1: Totally optional: Instead of abbreviating the Ext ending, maybe just write it out? `ApiFetcherSerializationExtensions`? (and `HttpFetcherSerializationExtensions`)?

Alternately, just `ApiFetcherExtensions` and `HttpFetcherExtensions` on its own is fine? The `Serialization` part is assumed because of the parent module we live inside.

These are minor comments and I don't feel too strongly about them.

settings.gradle.kts
```diff
@@ -27,6 +27,7 @@ include(":frontend:kobwebx-markdown")
 include(":frontend:compose-html-ext")
 include(":frontend:browser-ext")
 include(":frontend:test:compose-test-utils")
+include(":frontend:kobwebx-core-serialization")
```
bitspittle, line 30: No action required: This should be sorted, so I would personally would have put it after `:frontend:kobweb-worker-interface`). 

Looks like a few other modules are also out of order, however, which is why I'm not pushing you to fix the order with any urgency now. I can clean this up myself, in a followup commit.

My general advice (and advice can always be pushed back on or there can be exceptions!) is that code lists should always be ordered intentionally. The reason for this is simple -- if everyone on a large team is in the habit of just adding things to the end of every list, then it vastly increases the odds you have to deal with annoying merge errors when submitting code. Sorting things in some way makes it less likely that two programmers will stomp on the same line.

Ordered does NOT have to mean alphabetical -- in fact, alphabetical is a fallback when you can't think of anything better. It's totally fine to group things by behavior if that makes sense, for example. But at the very least, alphabetical is easy enough to understand and good enough in many cases.

bitspittle (commented):
Should be the only round of comments. I don't think anything should be too difficult to resolve!

frontend/kobweb-core/src/jsMain/kotlin/com/varabyte/kobweb/browser/ApiFetcher.kt
dead8309: But `window.api.get` don't accept `autoPrefix`. I think we dont need `Json.decodeFromString` as we are calling the http methods from `HttpFetcherSerializationExt.kt` which will handle json decoding

We want to prepend api calls with `/api` so i made that method public. Maybe i can add a private method in `ApiFetcherExtensions` file for that ?

bitspittle: The code snippet I included above is copied from my local copy of the code and it seems to be working fine? (`autoPrefix` is indeed an argument in `ApiFetcher.get`)

dead8309: Sorry!! I was reading that as window.http not window.api. making the change

dead8309:
Definately!!

dead8309:
We only want to support for serializing body for api routes right ? or do you think i should also add for http requests too

bitspittle:
> We only want to support body serialization for API routes, right? Or do you think I should also add support for HTTP requests as well?

Ah yeah we should do both. The APIs are definitely related and should feel consistent.

bitspittle (approved):
I'm going to go ahead and approve these changes. We'll merge after HttpFetcher also has its body APIs updated.

If you're busy and would like me to handle that in a followup commit, just let me know.

Thanks again for this PR!

PR #616, opened 2024-10-31
Disable compose trace markers for kobweb modules

DennisTsar:


bitspittle (approved):

bitspittle:
Looks great, thanks! Definitely appreciate the associated Gradle cleanup.

BTW I was using Kobweb and getting nearly instant refreshes when making changes, and really -- we wouldn't be here without your Gradle refactorings. I owe you big!

PR #626, opened 2024-12-12
Add configuration of compose compiler plugin to Kobweb gradle plugins

DennisTsar:


tools/gradle-plugins/core/src/main/kotlin/com/varabyte/kobweb/gradle/core/util/ComposeConfig.kt
```diff
@@ -0,0 +1,40 @@
...
+                    )
+                    null
+                }
+
+                else -> throw e
+            }
+        } ?: return@withPlugin
+        // Trace markers are "pure overhead" for the JS target & needlessly increase the bundle size, but
+        // must be explicitly disabled until https://youtrack.jetbrains.com/issue/KT-69900 is resolved
+        composeCompilerExt.includeTraceMarkers.set(false)
```
bitspittle, line 36: Should these calls should also be within a try / catch, just in case the user happens to be using a different compose plugin than the one we're expecting here?

Also, should both calls be in a single try/catch? Or one try/catch per setting?

tools/gradle-plugins/core/src/main/kotlin/com/varabyte/kobweb/gradle/core/util/ComposeConfig.kt
```diff
@@ -0,0 +1,40 @@
...
+        // Kobweb projects only use compose on the frontend, so we disable the compiler plugin for other targets
```
bitspittle, line 37: Optional: Unlike standard multiplatform applications, Kobweb projects only use ...

bitspittle (approved):

tools/gradle-plugins/core/src/main/kotlin/com/varabyte/kobweb/gradle/core/util/ComposeConfig.kt
DennisTsar: I couldn't find any explicit mention of a backwards compatibility policy, but some options deprecated in 2.0.20 are still around in 2.1.0. If it matches the policy of KGP, it would likely be several Kotlin versions before removal. Do you think it's worth adding the try/catch anyway to be safe?

bitspittle: Since our code handling is generally optional, I think a sloppy try/catch is fine here. I think it expresses the idea that: "I don't care *why* we fail, if we do it means we should just inform the user and abort. No harm no foul."

For context, I remember getting bit by https://github.com/varabyte/kobweb/blob/858378318eae1b492c30f4613ada3c2cdf435f1a/tools/gradle-plugins/application/src/main/kotlin/com/varabyte/kobweb/gradle/application/tasks/KobwebCopyTask.kt#L65

I think it's very very unlikely that your specific try/catch here will similarly blow up, so I'm certainly not going to block over this!

tools/gradle-plugins/core/src/main/kotlin/com/varabyte/kobweb/gradle/core/util/ComposeConfig.kt
```diff
@@ -0,0 +1,40 @@
...
+    pluginManager.withPlugin("org.jetbrains.kotlin.plugin.compose") {
+        // As described in https://github.com/gradle/gradle/issues/23792#issuecomment-1431053503,
+        // if the compose plugin is loaded in a separate classloader, we can't access and configure it.
+        val composeCompilerExt = try {
+            extensions.findByType<ComposeCompilerGradlePluginExtension>()
+        } catch (e: Throwable) {
+            when (e) {
+                is TypeNotPresentException, is ClassNotFoundException, is NoClassDefFoundError -> {
+                    project.logger.warn(
+                        "w: The Kobweb plugin failed to configure the compose compiler plugin. " +
```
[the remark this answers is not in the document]
bitspittle: As discussed offline, I think you came up with good wording here.

bitspittle (approved):

bitspittle (approved):

PR #629, opened 2024-12-15
Remove deprecations > 6 months old

DennisTsar:


bitspittle:
So satisfying ❤️  thank you!

PR #658, opened 2025-02-16
Fix popup reopening/staying open after switching browser tabs

DennisTsar:


playground/site/src/jsMain/kotlin/playground/pages/Widgets.kt
```diff
@@ -488,6 +490,15 @@ fun WidgetsPage() {
                         )
                     }
 
+                    Column(Modifier.gap(1.cssRem)) { // Should close after switching back to the tab
+                        Link(
+                            "https://example.com",
+                            "Click to open a new tab",
+                            openExternalLinksStrategy = OpenLinkStrategy.IN_NEW_TAB
+                        )
+                        Tooltip(ElementTarget.PreviousSibling, "tooltip", placement = PopupPlacement.Right)
```
bitspittle, line 499: Nit: How about a clearer tooltip, something like "This tooltip should not be open when you come back to this page."

DennisTsar: Good idea, done.

bitspittle (approved):

PR #663, opened 2025-03-02
Improve Callout widget spacing

DennisTsar:
This removes unwanted spacing from our docs site, which exposed some cases that were not being properly handled.


bitspittle (approved):
Thanks for looking into this and also updating the comment

PR #665, opened 2025-03-05
Remove legacy code/deprecations > 6 months old

DennisTsar:


bitspittle (approved):
Looks awesome!!

I may have screwed things up with a force commit, sorry. You may need to rebase and reupload. But you can then submit what you have -- this all looks good to me.

PR #668, opened 2025-03-12
Use Dokka to generate docs for this project and publish to GH Pages

### LINKED ISSUE #65: Use Dokka to generate docs for this project and publish to GH Pages

jakubmeysner:
- Created a new docs module used for Dokka multi-module documentation aggregation (see https://kotlinlang.org/docs/dokka-migration.html#update-documentation-aggregation-in-multi-module-projects)
  - Includes all compatible modules
- Added a GA publish docs workflow, that generates Dokka API documentation and publishes it to GH Pages

Demo: https://jakubmeysner.github.io/kobweb/

Resolves #65


settings.gradle.kts
```diff
@@ -14,6 +14,7 @@ include(":common:framework-annotations")
 include(":common:kobweb-common")
 include(":common:kobweb-serialization")
 include(":common:kobwebx-serialization-kotlinx")
+include(":docs")
```
bitspittle, line 17: We've actually used the `docs/` folder for our own docs in the past (e.g. migration docs). So I don't want to use that folder for a project...

I'm thinking `:tools:docs-generator` might be safer? I'll talk with a collaborator and confirm a final name for you soon.

.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,42 @@
...
+
+      - name: Setup Java
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: 11
+          cache: gradle
+
+      - name: Generate Dokka multi-module documentation
+        run: ./gradlew :docs:dokkaGenerate --no-daemon
```
bitspittle, line 33: Nit: Turns out adding --no-daemon is no longer a best practice. I may have some old legacy instances left over from my own scripts, in case you copy/pasted this from there, but we can safely remove it here.

.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,42 @@
+name: Publish Dokka API documentation
+
+on:
+  - workflow_call
+  - workflow_dispatch
```
bitspittle, line 5: This is the first time I've seen workflow_call. Any reason to do this with workflow_call instead of on publish? Feels like being more explicit and isolated might be better than having users realizing they should find this workflow called from another?

bitspittle (changes requested):
This is a really neat approach, thanks! 

Comments ahead but they should be relatively easy to resolve.

docs/build.gradle.kts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+group = "com.varabyte.kobweb"
+version = libs.versions.kobweb.get()
+
+dokka {
+    moduleName = "kobweb"
+}
+
+dependencies {
```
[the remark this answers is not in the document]
DennisTsar: This suggestion involves reaching into the internals of other projects, which is against Gradle best practices (not to mention it uses `Project.plugins`, which is soft-deprecated according to its docstring).

The explicit listing of projects is (unfortunately) the best practice.

.github/workflows/publish-docs.yml
jakubmeysner: @bitspittle it could theoretically be useful in case there was a need to publish the docs without running the full publish workflow (potentially when making changes to the Dokka configuration only?). However that's a rather obscure use case, so I can merge it into the publish workflow if you believe that there's no need to support it.

bitspittle: Oh, I definitely want this to be its own workflow script! I was just saying we should add the on published event in here and not as a workflow call. 

That said, I'm now thinking that keeping this as is is better, actually. So please ignore my initial comment. Apologies!

In the other script that calls this, we probably want to only call this one if we are publishing to sonatype. That way we'll avoid extra unnecessary publications when we push out the gradle plugins.

docs/build.gradle.kts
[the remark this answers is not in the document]
bitspittle: Ok Jakub you can leave this as is. I'll discuss the automated version with Dennis but it doesn't need to block your PR.

.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,42 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Java
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: 11
+          cache: gradle
```
DennisTsar, line 30: This should be replaced with the `gradle/actions/setup-gradle@v4` action that we use in the other workflows, which does more sophisticated caching.

bitspittle (approved):
Again, thank you for taking the time to help us out with this!

FYI, normally we enforce checking into the `dev` branch, although this is a rare commit that I would be pulling into main anyway (workflows are special and need to be in main to show up in the GitHub UI).

So you can submit into `main` with this one; just a heads up in case you ever want to help us out with future PRs.

P.S. I've already updated the settings for this repo to prepare for a GHP deployment from GitHub Actions.

settings.gradle.kts
jakubmeysner: @bitspittle one thing that I'm not really sure about is the `generator` part, since the module doesn't do the docs generation, but rather aggregates the docs from all the different modules in its own docs (both `:some-module` and `:docs` have the `dokkaGenerate` tasks, but `:docs` pulls in the docs from other modules), so if I were to suggest something it'd be skipping that part and rather going with something like `:tools:docs`, `:tools:all-docs`, `:tools:api-docs`, `:tools:all-api-docs`, or similar.

bitspittle: `:tools:aggregate-docs` maybe?

bitspittle (approved):
Thank you again! I'll pull this in now.

PR #671, opened 2025-03-18
Make invalid CssStyle error message lazy

DennisTsar:
This avoids an unnecessary performance penalty from `trimMargin()` calls that never get used.


bitspittle (approved):
Nice, thanks.

PR #674, opened 2025-03-20
Add support for repeating gradients

bitspittle:


bitspittle:
When playing around with your new gradient APIs, I saw a repeating gradient example and thought I'd look into seeing if we could add support for that too. It turns out, the repeating APIs take the exact same arguments as the underlying non-repeating types. Easy!

I believe repeating gradients don't make sense without explicit color stops though -- so I don't have two-color convenience versions of them.

Code is pretty trivial but I could be missing something so a separate set of eyes would be great here. Not urgent.

DennisTsar (approved):
Looks good!

For consistency with the other gradients, `RepeatingGradient` should have an `internal` constructor. The delegation also isn't doing anything, though I like the display of intention.

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/functions/Gradients.kt
```diff
@@ -44,6 +44,12 @@ internal typealias LengthColorStopsBuilderEntry = Gradient.ColorStopsBuilder.Ent
 typealias AngleColorStopsBuilder = Gradient.ColorStopsBuilder<CSSAngleNumericValue>
 internal typealias AngleColorStopsBuilderEntry = Gradient.ColorStopsBuilder.Entry<CSSAngleNumericValue>
 
+// All repeating gradients take the same arguments as the non-repeating versions; they just use a different name and
+// interpret the arguments differently.
+class RepeatingGradient<G: Gradient>(private val wrapped: G) : Gradient by wrapped {
```
DennisTsar, line 49:
```suggestion
class RepeatingGradient<G : Gradient> internal constructor(private val wrapped: G) : Gradient by wrapped {
```

bitspittle:
(Ended up submitting directly to dev)

PR #675, opened 2025-03-22
Switch to `@scope`-based color mode implementation

DennisTsar:


frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Scope.kt
```diff
@@ -0,0 +1,54 @@
...
+ *                 backgroundColor(Black)
+ *             }
+ *         }
+ *     }
+ * }
+ * ```
+ */
+fun <TBuilder> GenericStyleSheetBuilder<TBuilder>.scope(
+    start: String?,
+    end: String?,
```
bitspittle, line 49: Should `start` and `end` default to null?

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Scope.kt
```diff
@@ -0,0 +1,54 @@
...
+    rulesBuild: GenericStyleSheetBuilder<CSSStyleRuleBuilder>.() -> Unit
```
bitspittle, line 50: (I'm assuming you copied this from somewhere else, but I wonder if this should be called `rulesBuilder` or `buildRules`?)

frontend/silk-foundation/src/jsMain/kotlin/com/varabyte/kobweb/silk/style/ColorModeStrategy.kt
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * Indicates whether color mode aware styles should be registered using `_light` and `_dark` suffixes.
+ *
+ * Note that [useScope] and [useSuffix] are NOT mutually exclusive.
+ */
+val ColorModeStrategy.useSuffix get() = this != SCOPE
+
+/**
+ * Returns a list of CSS layers that will be used by the site, based on the given strategy.
+ *
```
bitspittle, line 90: Optional: It might be worth mentioning on extra line here (unless you left this comment elsewhere and I missed it):

Color-aware styled are bound to different CSS layers, based on which color mode strateg we're using.

frontend/silk-foundation/src/jsMain/kotlin/com/varabyte/kobweb/silk/style/CssStyle.kt
```diff
@@ -216,6 +217,43 @@ abstract class CssStyle<K : CssKind> internal constructor(
...
+     */
+    private fun withColorModeScope(
+        selector: String,
+        group: StyleGroup,
+        baseStyleSheet: StyleSheet,
+        lightStyleSheet: StyleSheet,
+        darkStyleSheet: StyleSheet,
+        handler: GenericStyleSheetBuilder<CSSStyleRuleBuilder>.(String, ComparableStyleScope) -> Unit
+    ) {
+        // A selector like ".abc" inside an `@scope` rule is implicitly treated as ":scope .abc", which means that it
```
bitspittle, line 234: No action required: This is an amazing comment. Write stuff like this at work and developers will cry in awe when they stumble upon it.

bitspittle (approved):
Looks faaantastic. I can tell the work that went into this behind the scenes.

I pulled the code down and tested a production export on chrome, firefox, and safari. FYI I still get the flash on Chrome (and Firefox of course), but Safari seems to have worked.

Still, everything is functional, so I'm going to go ahead and approve. If you can go ahead and do a production export on Chrome and set the network tab to throttling 3G or something and refresh, let me know what you see?

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Scope.kt
DennisTsar: They could, but most of the time you want at least one to not be null, so I'm not sure if making `scope {}` work is beneficial.

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Scope.kt
DennisTsar: Indeed, I copied it from `Layers.kt`, which in turn is copied [JB](https://github.com/JetBrains/compose-multiplatform/blob/3c0081d46678318c8c68c972987c599ebc754164/html/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/CSSMediaRule.kt#L78), which uses it like `buildRules(rulesBuild)`. It's not clear to me why our `layer()` implementation doesn't do it that way as well.

frontend/silk-foundation/src/jsMain/kotlin/com/varabyte/kobweb/silk/style/ColorModeStrategy.kt
DennisTsar: The docstring says
> When the strategy is [BOTH], each original layer is preceded by a corresponding "compatibility" layer. These compatibility layers contain suffixed styles that also exist as `@scope`-based styles, meaning all styles in these
layers can be ignored if the browser supports `@scope`.

Does that cover your request?

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Scope.kt
bitspittle: According to the docs for scope block, even empty scopes have a use. But also, I think I'd want to specific an end scope only without a start one, and now I'd have to add an awkward `null` as a first parameter, instead of just writing `scope(end = ...) { ... }`

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Scope.kt
bitspittle: I'm ok if we change it in both places but that could be a follow-up commit and I don't really care that much!

bitspittle (approved):

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Scope.kt
DennisTsar: Ok, I've made the change.

bitspittle (approved):
Commit at will. This feels kind a monumental change in the history of the Kobweb codebase 🫡

Issue #611, opened 2024-09-17
Remove .idea/ from version control


okarmazin:
`.idea/` should not be checked in to version control since it contains local installation state. There may be the rare shareable configuration, but the vast majority of the content is local to the current IDE installation.

Case in point, just opening Kobweb in my IDEA installation modifies some existing configuration files in `.idea/`

Kobweb's `.gitignore` rules allow local IDE state leakage.

I propose that all `.idea/` derectories be removed from VCS entirely.

bitspittle:
I totally understand where you're coming from but I disagree with the strength of the opinion. I should actually update my CONTRIBUTING doc to be clear about this. 

In short: if people are contributing to Kobweb, I strongly prefer they use IntelliJ IDEA to do so. There is a lot of junk in the .idea that is optional (and we can exclude more of it if identified), but not all of it -- we have been adding (and are open to add more) various formatting rules in there.

If we get more and more contributors over time, it's going to be hard enough to review PRs without some standards, and the .idea folder gives us some tools to help at least remove some of the potential noise.

Here is, I believe, official guidance from JetBrains: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839-How-to-manage-projects-under-Version-Control-Systems

lppedd:
The problem with sharing `.idea` is that configuration files change between IDEA releases.  
I do commit a subset of the folder usually, but only if I'm sure the team is mostly synced with IDE versions.

bitspittle:
(Sorry for no reply last week -- was dealing with a bunch of real life distractions!)

Just wanted to thank @lppedd for the extra context. I am going to think about it.

I am also wondering if in general I should be encouraging developers to use either the latest stable or EAP version of IntelliJ IDEA. Of course, if someone uses their own IDE and submits a PR, I won't reject it; but by default, especially if a contributing dev doesn't really have any strong opinions about their own workflow, I'd like to feel like I can rely on specifying some syntax rules in the IDE settings with a chance that uesrs will have them applied in their own code for consistency.

EchoEllet:
> In short: if people are contributing to Kobweb, I strongly prefer they use IntelliJ IDEA to do so. There is a lot of junk in the .idea that is optional (and we can exclude more of it if identified), but not all of it -- we have been adding (and are open to add more) various formatting rules in there.
> 
> If we get more and more contributors over time, it's going to be hard enough to review PRs without some standards, and the .idea folder gives us some tools to help at least remove some of the potential noise.

[...]

PR #679, opened 2025-03-28
Removed .idea/ from version control

vrajatcreations:
Removed .idea/ from version control

#611


bitspittle:
I'm going to reject this because we add the .idea/ folder on purpose. We expect most contributors will use IDEA, and we include a handful of settings in to help them (as mentioned in the bug that you linked to)

If there's any specific file that got checked into the IDEA folder that you think shouldn't have been checked in, please let me know.

(That bug you linked to was opened by a third party so maybe you saw it and thought it would be an easy fix, but I never really agreed to the issue in the first place. I'll update its wording to prevent confusion in the future)

PR #680, opened 2025-03-28
Changes regarding converting base class

vrajatcreations:
Changes regarding converting base class Done for only two Classes, is this approach is correct, do in other files also


bitspittle:
Just a heads up that last week's release took way longer and much more work than we expected, plus these last two days have needed a bunch of my time for personal chores, and I have a potentially big bug with the last release I'll be looking into next (which might require new releases). This PR is top priority after that. 

Apologies for the delay -- I did not forget your PR and I have a daily reminder that will ping me forever until I finally am able to get to a review.

bitspittle (changes requested):
Hello! Again, sorry so much for the delay here. This was hopefully an example of worst case timing and it shouldn't happen again.

*Quick note: please submit PRs against the `dev` branch (you may want to review the [CONTRIBUTING.md](https://github.com/varabyte/kobweb/blob/main/COMPATIBILITY.md) document, it's pretty short). In your local fork, please grab latest, rebase against dev, and then push up your PR again later, thank you!*

And finally! The review: I like the initial idea here behind extracting a generic parameter, but the approach needs a little work. It's very important that the keywords are companion object properties.

In other words, a user should write `Animation.Inherit` but this should NOT work: `Animation.of(...).Inherit`.

---

I *think* you can accomplish what you are going for with this pattern:

```
interface CssGlobalValues<T : StylePropertyValue> {
    val Inherit: T get() = "inherit".unsafeCast<T>()
    val Initial: T get() = "initial".unsafeCast<T>()
    val Revert: T get() = "revert".unsafeCast<T>()
    val Unset: T get() = "unset".unsafeCast<T>()
}

sealed class Animation private constructor(value: String) : StylePropertyValue {
   companion object : CssGlobalValues<Animation>
        // Keyword
        val None: Animation get() = Keyword("none")
    }
}

sealed class BackgroundAttachment private constructor(value: String) : StylePropertyValue {
   companion object : CssGlobalValues<Animation>
        // Keyword
        val Scroll get() = BackgroundAttachment("scroll")
        val Fixed get() = BackgroundAttachment("fixed")
        val Local get() = BackgroundAttachment("local")    
    }
}
```

Can you check if that works and reupload the PR?

vrajatcreations:
@bitspittle Thanks for your patience in explaining the issue and for giving me the opportunity to contribute to this great project. I've made the changes as discussed and tested the code in the **RawHTML** file — attaching a screenshot for reference.

Screenshot 2025-04-12 at 7 42 48 PM



If this looks good, I'll go ahead and apply the changes across all relevant files

bitspittle:
Thank you @vrajatcreations .

But how come you didn't use the format [suggested in my comment above](https://github.com/varabyte/kobweb/pull/680#pullrequestreview-2754928075)? When I tested it locally, it worked, and it's a little easier to type than putting `by` everywhere.

---

Also, heads up I will need to discuss this approach with the other major Kobweb contributor. There is one thing I am a bit worried about is people copy/pasting the wrong class in the generic type without realizing it.

For example, did you notice the copy/paste error I intentionally added in my comment above? It's really hard to spot.

sandeepjak2007:
@bitspittle will follow the format which you mentioned and re raise the pr once again.

DennisTsar:
@bitspittle Actually I think the current approach with `globalStyleValues(factory)` is more type safe. It also helps avoid passing the wrong generic since private constructors of other classes aren't accessible when passed explicitly. 

One thing to watch out for with this approach is that sealed classes must reference a concrete constructor, so for example `BackgroundPosition` should use `globalStyleValues(BackgroundPosition::Keyword)` instead of `::BackgroundPosition`.

sandeepjak2007:
@bitspittle @DennisTsar 

Thank you both for your patience and guidance throughout this contribution. I've made the requested changes.

Please close this PR and I've opened a [new one](https://github.com/varabyte/kobweb/pull/690) with the updates.

Please take a look and let me know your thoughts!

bitspittle:
Thank you for opening a new PR to clean up the history.

Note that in the future I think (not sure) you can also do a `git commit --amend` followed by a `git push --force` which should update the PR in place.

bitspittle:
This fix went in as: https://github.com/varabyte/kobweb/commit/418094a2ed45fcdd54f4378f69bd3baddced0543

Issue #175, opened 2022-11-04
Update docs with "Link" -> "Anchor" renaming.


bitspittle:
The relevant location:

- [x] README
- [ ] Code comments
- [ ] Website docs (https://kobweb.varabyte.com)

bitspittle:
Looking back at this bug, I can't remember what it was referring to exactly. Probably part of the README that was deleted a long time ago.

PR #689, opened 2025-04-12
Update docs with "Link" -> "Anchor" 

sandeepjak2007:
Update docs with "Link" -> "Anchor" by renaming


bitspittle:
Hey, apologies, I don't remember why I filed issue #175 or what it was referring to. Probably something that got removed a long time ago.

If you check out the official site, you'll see that [we actually use `Link` in the hero example](https://github.com/varabyte/kobweb-site/blob/4963491d5fc212e7ef2baaea9f9ed04caba11021/site/src/jsMain/kotlin/com/varabyte/kobweb/site/components/sections/home/HeroSection.kt#L120), and honestly that's on purpose. Most users should be using `Link`, not `Anchor`.

Sorry if having that random old issue around wasted your time. In the future, please consider leaving a comment on a bug first before starting work on it so I can save you from working on something that shouldn't actually be open.

PR #690, opened 2025-04-13
Changes regarding converting base class

sandeepjak2007:
Please check the PR and let me know if any corrections

### DIFF
--- a/frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Alignment.kt
+++ b/frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Alignment.kt
@@ -2,6 +2,7 @@
 
 package com.varabyte.kobweb.compose.css
 
+import com.varabyte.kobweb.compose.css.global.CssGlobalValues
 import org.jetbrains.compose.web.css.*
 
 // !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
@@ -62,7 +63,7 @@ sealed class AlignContent(private val value: String) : StylePropertyValue {
     private class OverflowAlignment(strategy: OverflowStrategy, position: AlignContentPosition) :
         AlignContent(strategy.toValue(position))
 
-    companion object {
+    companion object: CssGlobalValues<AlignContentKeyword> {
         // Basic
         val Normal: AlignContent get() = AlignContentKeyword("normal")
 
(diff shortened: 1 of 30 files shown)


bitspittle:
(It's late Sunday over here now; I'll review this tomorrow!)

bitspittle:
Hey, you've already done so much work and been patient with us. We're definitely going to submit something and you'll get credit for it -- and definitely our thanks.

Still, your change has made me realize that we still need to do a bit more. I'll talk more about that later (*).

I'd like to ask you how like to proceed here. I am happy to go one of two ways:

1) take this commit as is and then our team will tweak it further in followup commits.
2) be willing to go back and forth with us a bit more, but this might take a while, especially we're in different time zones it seems.

At first, I was going to just push for 1), but I want to make sure you feel appreciated! And maybe this is a learning experience for you, and maybe you'd feel frustrated if we took your work from you. So I thought instead of making a decision myself, I would ask you what you prefer.

---

(*) As for what extra needs to be done...

* First, I think you may have missed this comment: https://github.com/varabyte/kobweb/pull/680#issuecomment-2800046390. I feel really bad because Dennis actually pointed out why your original way was better (it is extra typing but more type safe)
* Second, I think we should be adding tests for all these properties. This will be a lot of work and it's something I can do; I don't want to ask you or have you feel obligated at all to do it.

sandeepjak2007:
@bitspittle Thanks for your patience and thoughtful feedback.

As you suggested, I’ll prioritize **Option 1** and would love to start with the first point you marked with (*). It would be great if you could guide me through the changes, especially since we're in different time zones and it's important to keep things moving.

Also, if there are any smaller or simpler changes that you or the team are working on, I’d be happy to help out with those too. I’m still learning my way around this large project and would really appreciate the opportunity to contribute in small ways and grow with the team.

bitspittle:
Thank you for the quick reply. Just to make sure, you are OK with **option 1** which is where the team will take over (but keep you in the loop). Please double check and let me know if you meant **option 2**.

Otherwise, my plan actually is to prioritize writing tests for all CSS propeties, get that working first, and then, once done, layer your change on top of those tests. (If there are no problems with your changes, then all the tests will still pass). And finally, look into migrating the advice left by Dennis on the other PR. So three PRs in total.

As we're getting closer to 1.0, there's really two main high level things we have to do at this point -- add more widgets (this is really tricky and not something I'd recommend) and finish binding all CSS APIs with Kotlin code. This can also be kind of tricky, but there's a lot of existing sample code to learn from.

So if you're interested, [check out this spreadsheet](https://docs.google.com/spreadsheets/d/1Uu2diibyOzDFPgzzM8BWlw4B9_CTDBACv_1E_tsXbp8), find a CSS property still unchecked, look up the official docs for that property (try to find ones that look really easy, at first), and then take a stab and writing Kotlin code for it. Especially now you are initially familiar with where our CSS code lives.

---

Finally, note that I'm in the middle of finishing up another feature. So it will be another two days or so before I can start implementing the tests for these CSS properties.

sandeepjak2007:
@bitspittle 

For now, I'll proceed with **Option 1**. Please let me in loop so that learn it.

Also, for the unchecked CSS properties, I’d be happy to pick one and try making small changes to get started. Could you let me know how you'd prefer me to showcase those small updates for your review?

bitspittle:
Just go through the CSS list, choose one item that seems simple, and upload a PR for it. Hopefully there are still a bunch of easy ones left in the list, I'm not sure.

Please don't feel any rush to get a PR up. It can take a while to learn code, and that's normal. You can ask me questions about it if you have any.

bitspittle:
This was submitted at https://github.com/varabyte/kobweb/commit/418094a2ed45fcdd54f4378f69bd3baddced0543, thank you!

PR #691, opened 2025-04-15
Css checklist for break properties

sandeepjak2007:


bitspittle:
Thank you for the PR. It's been a crazy day and tomorrow I have a bunch of personal chores to do as well, but I will take time soon to get more familiar with the CSS break properties and hopefully can get back to you within two days.

sandeepjak2007:
No worries at all, take your time! Thanks for the update — I’ll wait to hear back from you.

bitspittle:
Just wanted to let you know I'm on track to review this tomorrow.

sandeepjak2007:
Thanks for the update! Please review and let me know if I’m heading in the right direction. If not, I’d appreciate your guidance so I can contribute more effectively—fully understand that maintaining an open source project is a big effort and I want to respect your time.

bitspittle:
Thanks for understanding. We're just caught on a feature taking longer to get in that originally thought, ideally before the KotlinConf talk coming up (which I haven't been able to really even start yet), so it's been stressful. I'm getting too old to be doing 14 hour coding days, day after day, but here we are...!

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Break.kt
bitspittle, line 1: I appreciate the attempt to use an interface to refactor out common properties, but I worry it feels a bit too abstract now. Like, even me as the owner of the project is looking at the code and finding it hard to read. Compare that to all the other CSS properties, which you can see at a glance what their list of properties are.

The reason we want to extract the global keywords into a shared interface is because, well, they're global. They apply to *everything*. It does add some complexity, which is unfortunate, but because it's so widely applicable it's probably worth the sacrifice.

But I'm not sure how much of a win it is here.

I can see an argument for refactoring out the properties for `break-before` and `break-after` -- they're clearly the same and will always be kept in sync. So *maybe* we could extract an interface for these values eventually...

```css
/* Generic break values */
break-after: auto;
break-after: avoid;
break-after: always;
break-after: all;

/* Page break values */
break-after: avoid-page;
break-after: page;
break-after: left;
break-after: right;
break-after: recto;
break-after: verso;

/* Column break values */
break-after: avoid-column;
break-after: column;

/* Region break values */
break-after: avoid-region;
break-after: region;
```

but, honestly, can we keep it simple for a first pass?

Go ahead and repeat the code for now -- for example, look at my implementations of `com.varabyte.kobweb.compose.css.Width` and `.Height` -- That will be more consistent with the rest of the codebase and will be easier for me to review. For example, I'm pretty sure you don't need to use a sealed class + `Keyword` pattern here but because of your interface approach I can't tell.

And definitely please don't extract the global properties into a separate interface, since as you know, we are going to be doing that separately soon.

bitspittle (changes requested):
Sorry again it took me so long to take a look at this.

BTW these were great CSS properties to start with. Definitely it's nice when you can find CSS properties that are just a list of keywords.

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Break.kt
sandeepjak2007: Thank you. Again thanks for your guidance with patience. Will rework according to mentioned classes and push the changes

bitspittle: You're very kind. You don't have to be so polite with me -- after all, you're doing me a big favor too. I appreciate it.

I also know code reviews can be annoying, especially maybe at some point you might disagree with me but not want to say it. But you are always welcome to say it, just so you know.

You've been very patient with me so you can bet I'll return the favor and be patient with you too.

sandeepjak2007: Implemented the suggested changes — I believe I understood your comment correctly and updated the code accordingly. Please review it when you get a chance and let me know if any further adjustments are needed.

bitspittle (approved):
Looks perfect, thank you!

Submitting shortly.

One day we can review all properties and, like BreakAfter and BreakBefore, and Width and Height, decide if we should merge some of them somehow.

bitspittle:
And thank you @sandeepjak2007 

Just as a heads up, my plan is in the next two or three days to write tests for CSS properties and then do the global keywords refactoring. I'll be sure to ping you when those go in.

PR #692, opened 2025-04-22
Modifier support for column-count, column-fill, column-span CSS property

sandeepjak2007:
I've added support for the CSS properties `column-count`,` column-fill,` and `column-span`.
Please take a look and let me know if any changes are needed. Thanks!

### DIFF
--- a/frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Column.kt
+++ b/frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Column.kt
@@ -1,0 +1,72 @@
+package com.varabyte.kobweb.compose.css
+
+import org.jetbrains.compose.web.css.*
+
+// https://developer.mozilla.org/en-US/docs/Web/CSS/column-count
+
+class ColumnCount private constructor(private val value: String) : StylePropertyValue {
+
+    companion object {
+
+        /* Keyword value */
+        val Auto get() = ColumnCount("auto")
+
+        /* <integer> value */
+        fun of(count: Number) = ColumnCount("$count")
+
+        /* Global values */
+        val Inherit get() = ColumnCount("inherit")
+        val Initial get() = ColumnCount("initial")
+        val Revert get() = ColumnCount("revert")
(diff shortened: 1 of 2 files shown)


bitspittle:
(Ignore the failing tests, that looks like something we did)

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Column.kt
```diff
@@ -0,0 +1,72 @@
...
+        /* Global values */
+        val Inherit get() = ColumnCount("inherit")
+        val Initial get() = ColumnCount("initial")
+        val Revert get() = ColumnCount("revert")
+        val RevertLayer get() = ColumnCount("revert-layer")
+        val Unset get() = ColumnCount("unset")
+    }
+}
+
+fun StyleScope.columnCount(columnCount: ColumnCount) {
```
bitspittle, line 26: 99% of the time it will be a count value, so let's go ahead and add `fun StyleScope.columnCount(count: Int)` as well (and a corresponding modifier)

bitspittle (changes requested):
Very minor changes but after that this is definitely approved.

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Column.kt
DennisTsar: We've previously discussed adding such convenience functions only for `Modifier`, not `StyleScope`. So I would propose justhaving `fun Modifier.columnCount(count: Int) = this.columnCount(ColumnCount.of(count))`. Or do you think we should reconsider that rule?

bitspittle (approved):
Thank you!

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/css/Column.kt
bitspittle: Ah no you're totally right. It's just been a while and I forgot. Thank you.

I'll remove the stylescope method but leave the modifier one.

PR #703, opened 2025-06-06
Fix live reloading when multiple sites exist in the project

DennisTsar:


tools/gradle-plugins/application/src/main/kotlin/com/varabyte/kobweb/gradle/application/buildservices/KobwebTaskListener.kt
```diff
@@ -16,15 +16,17 @@ import java.io.File
...
+         * A map of fully qualified paths of `kobwebStart` tasks with their corresponding Kobweb folder.
+         *
+         * This should only contain tasks that will run in the current build.
+         */
+        var kobwebFolderFiles: Map<String, File>
     }
 
-    val kobwebFolder = KobwebFolder(parameters.kobwebFolderFile.toPath())
-    var isBuilding = false
+    private val kobwebFolders = parameters.kobwebFolderFiles.mapValues { (_, file) -> KobwebFolder(file.toPath()) }
```
bitspittle, line 28: This might be less confusing if we introduce a named class? Something like

data class KobwebStartTaskInfo(val startTask: String, val folder: KobwebFolder)

Otherwise, I'm having trouble thinking of a name for it. It's not really "kobwebFolders" -- maybe "kobwebStartTaskFolderMapping" but that's a mouthful?

bitspittle (commented):
Overall, I'm definitely OK with this. Just having some issues with naming, and also there should maybe be a comment somewhere explaining why we're doing all this tricky work? (Something simple like "A user's project may have multiple sites in their project, with multiple sites potentially running at the same time. We want to ensure that live reloading works for all of them.")

DennisTsar:
Applied your suggestions and added some clarification, let e know if anything is still confusing.

I'll also note that we could avoid the complexity by just not supporting live reloading for multiple sites simultaneously and printing a warning if we detect the user attempting to do so. The cli doesn't support running multiple sites at once anyway, and this still causes all sites to reload even when only one of them is changed.

bitspittle (approved):
Looks good! I like the names of everything.

PR #707, opened 2025-06-14
fix: `ArrayBuffer`, `MessagePort`, and `ImageBitmap` being unretrievable from `Transferables`

shaksternano:
Changes `suffixedKey` to not append an underscore when `suffix` is empty, fixing the key mismatch when adding and retrieving  `ArrayBuffer`, `MessagePort`, and `ImageBitmap` to and from `Transferables`.


bitspittle:
I looked at the code and I think the right fix is, instead, to change this line?

Transferables.kt, L123
> private fun _add(key: String, value: Any) = _add(key, suffix = null, value)

shaksternano:
> I looked at the code and I think the right fix is, instead, to change this line?
> 
> Transferables.kt, L123 from:
> 
> > private fun _add(key: String, value: Any) = _add(key, suffix = "", value)
> 
> to:
> 
> > private fun _add(key: String, value: Any) = _add(key, suffix = null, value)

Done

bitspittle (approved):

PR #724, opened 2025-07-23
Preserve property order and repeated declarations in `ComparableStyleScope`

DennisTsar:
This fixes an issue where the style
```kt
val SomeStyle = CssStyle.base { Modifier.margin(10.vw).margin(10.dvw) }
```
would be registered as `.some { margin: 10dvw; }`, ignoring the fallback declaration and breaking styling for older browsers that do not support `dvw`.

This also brings the behavior of `CssStyle` in line with inline `Modifier` styles, which already preserved repeated styles.


bitspittle (approved):
Thank you for this commit, and all the prior investigation and work to get here.

The only comment I thought about leaving and then yanked was extracting the duplicated "properties.joinToString(...)" stuff into an extension method on top of `Collection<CssPropertyDeclaration>` but I ultimately talked myself out of it. Still, wanted to mention it here for some permanent record of my thoughts.

PR #737, opened 2025-08-21
Image loading attributes

DatL4g:
Added loading attributes for images which are pretty important.


bitspittle (changes requested):
Thank you for extending the image class with these useful attributes!

bitspittle (approved):
Looks great, thanks again!

I'll kick off a snapshow after merging which should be available in about 20-30 minutes.

PR #743, opened 2025-09-02
Fix main template `basePath` handling

DennisTsar:
`BasePath.remove` requires an absolute path


bitspittle (approved):
Tested, looks good!

PR #757, opened 2025-10-22
Delete >6 month deprecations

DennisTsar:


bitspittle (approved):
I can't believe the background blend changes were more than a month ago! 😅

PR #760, opened 2025-11-19
Migrate away from `kotlin-dsl` plugin

DennisTsar:


bitspittle (commented):
Wow, a bit surprised things changed so much around the java gradle plugin. I guess I should look into 2.3.0 :)

Is the intention for this to be a commit preview, and we wait until after the RC is out to update it?

build.gradle.kts
```diff
@@ -48,4 +55,30 @@ subprojects {
             compilerOptions.freeCompilerArgs.add("-Xjdk-release=${jvmTarget.target}")
         }
     }
+
+    pluginManager.withPlugin("java-gradle-plugin") {
```
bitspittle, line 61: Is it worth leaving some kind of high level explanatory comment? Otherwise I think people might be surprised to see reference to java-gradle-plugin support in a Kotlin project.

Maybe something as simple as:

// The following configuration exists since we stopped applying the kotlin-dsl plugin
// See also: https://mbonnin.net/2025-07-10_the_case_against_kotlin_dsl/

bitspittle (approved):

PR #764, opened 2025-12-17
Enable the unused return value checker

DennisTsar:
In the process, mark ignorable functions and adjust some functions to no longer return unnecessary values.

Notably, this also resolves https://github.com/varabyte/kobweb-intellij-plugin/issues/4 (for users that enable the feature).

Note that while the checker is experimental, it is safe to use in libraries:
> You can freely switch between compiler modes. Despite the feature being experimental, Kotlin metadata for it is compatible in both ways, and using Full mode does not force the compiler to emit pre-release binaries. Compiling your library in Full mode does not cause any problems for users with the feature disabled.

See: https://github.com/Kotlin/KEEP/blob/main/proposals/KEEP-0412-unused-return-value-checker.md#opting-in-your-library

### DIFF
--- a/common/kobwebx-frontmatter/src/commonTest/kotlin/com/varabyte/kobwebx/markdown/FrontMatterElementTest.kt
+++ b/common/kobwebx-frontmatter/src/commonTest/kotlin/com/varabyte/kobwebx/markdown/FrontMatterElementTest.kt
@@ -56,10 +56,10 @@ class FrontMatterElementTest {
         assertThat(fmElement["data.assets.font.Roboto"]).isNull()
 
         // Ensure getValue throws or not throws (as expected)
-        fmElement.getValue("title")
-        fmElement.getValue("data.assets.images")
-        fmElement.getValue("data.assets.images.0")
-        fmElement.getValue("data.assets.font")
+        val _ = fmElement.getValue("title")
+        val _ = fmElement.getValue("data.assets.images")
+        val _ = fmElement.getValue("data.assets.images.0")
+        val _ = fmElement.getValue("data.assets.font")
         assertThrows<NoSuchElementException> { fmElement.getValue("data") }
         assertThrows<NoSuchElementException> { fmElement.getValue("data.x.y.z") }
         assertThrows<NoSuchElementException> { fmElement.getValue("data.assets.images.5") }
@@ -87,7 +87,7 @@ class FrontMatterElementTest {
         }
 
         // Values are fine
-        FrontMatterElement.Builder {
+        val _ = FrontMatterElement.Builder {
             addScalar("validKey1", "valid.value")
             addList("validKey2") {
                 addScalar("valid.value")
--- a/frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/ui/Modifier.kt
+++ b/frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/ui/Modifier.kt
@@ -72,6 +72,7 @@ fun Modifier.thenUnless(condition: Boolean, other: Modifier): Modifier {
  */
 @OptIn(ExperimentalContracts::class, ExperimentalExtendedContracts::class)
 inline fun Modifier.thenIf(condition: Boolean, lazyProduce: () -> Modifier): Modifier {
+    @Suppress("RETURN_VALUE_NOT_USED") // `holdsIn` is incorrectly not marked as ignorable
     contract { condition holdsIn lazyProduce }
     return this.then(if (condition) lazyProduce() else Modifier)
 }
@@ -88,6 +89,7 @@ inline fun <T> Modifier.thenIfNotNull(value: T?, consume: (T) -> Modifier): Modi
  */
 @OptIn(ExperimentalContracts::class, ExperimentalExtendedContracts::class)
 inline fun Modifier.thenUnless(condition: Boolean, lazyProduce: () -> Modifier): Modifier {
+    @Suppress("RETURN_VALUE_NOT_USED") // `holdsIn` is incorrectly not marked as ignorable
     contract { !condition holdsIn lazyProduce }
     return this.thenIf(!condition, lazyProduce)
 }
--- a/frontend/silk-foundation/src/jsMain/kotlin/com/varabyte/kobweb/silk/style/CssStyle.kt
+++ b/frontend/silk-foundation/src/jsMain/kotlin/com/varabyte/kobweb/silk/style/CssStyle.kt
@@ -435,8 +435,10 @@ internal class SimpleCssStyle(
     extraModifier: @Composable () -> Modifier,
     val layer: String?
 ) : CssStyle<GeneralKind>(init, extraModifier) {
-    internal fun addStylesInto(styleSheet: StyleSheet): ClassSelectors {
-        return addStylesInto(selector, styleSheet, layer)
+    internal fun addStylesInto(styleSheet: StyleSheet) {
+        // Ignore the returned selectors as `SimpleCssStyle` wraps `stylesheet.registerStyle(...)` rules which are
+        // never applied to an element via `toModifier()`.
+        val _ = addStylesInto(selector, styleSheet, layer)
     }
 }
 
--- a/gradle/build-logic/publish/src/main/kotlin/com/varabyte/kobweb/gradle/publish/KobwebPublishPlugin.kt
+++ b/gradle/build-logic/publish/src/main/kotlin/com/varabyte/kobweb/gradle/publish/KobwebPublishPlugin.kt
@@ -9,8 +9,8 @@ import org.gradle.api.tasks.TaskContainer
 import org.gradle.jvm.tasks.Jar
 import org.gradle.kotlin.dsl.create
 import org.gradle.kotlin.dsl.named
-import org.gradle.kotlin.dsl.register
-import org.jetbrains.dokka.gradle.tasks.DokkaGenerateTask
+import org.jetbrains.kotlin.gradle.dsl.HasConfigurableKotlinCompilerOptions
+import org.jetbrains.kotlin.gradle.dsl.kotlinExtension
 import javax.inject.Inject
 
 abstract class KobwebPublicationConfig @Inject constructor(objects: ObjectFactory) {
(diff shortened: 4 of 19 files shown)


backend/io/src/test/kotlin/com/varabyte/kobweb/io/ByteSourceTest.kt
```diff
@@ -72,6 +72,7 @@ class ByteSourceTest {
     }
 
     @Test
+    @Suppress("RETURN_VALUE_NOT_USED") // Perhaps the `assertThrows` lambda should return `Any` instead?
```
bitspittle: Can you explain this more? Why isn't `Unit` better? Could we not use the `@IgnorableReturnValue` annotation instead?

bitspittle (commented):
Feels like I should fix up truthish first? We could remove a bunch of "val _" in that case, right? I think I can throw out a 1.0.4 in the next day or two

backend/io/src/test/kotlin/com/varabyte/kobweb/io/ByteSourceTest.kt
DennisTsar: There are two values at play here. Adding `@IgnorableReturnValue` to `assertThrows` would say that the `T : Throwable` that it returns is ignorable, which would be correct, but is not the issue here (since truthish has not opted-in to the return value checker so its functions are ignored for the inspection).

The warning here comes from the fact that the result of the lambda is never used, which is correct since it's coerced to `Unit`, and the only way to fix it is to either explicitly ignore it within the lambda (`val _ =`) or make it used (change the lambda to "eat" the returned value with return type `Any?`). Beyond just pleasing the checker, It feels more conceptually correct to say that `assertThrows` is meant to capture values rather than just ignore them - it's just that the value isn't actually expected be returned.

`kotlin.test.assertFailsWith` (which actually works as a drop-in replacement for `assertThrows` here) made this [same change](https://github.com/JetBrains/kotlin/commit/bd479100f528af3d7631577a71b2771ea753adc8#diff-f47416122ee7e1f687caaa7bf117a241de02e3c7d51d1afd08bd0b23545ea7a3R665).

bitspittle: Thanks for the explanation! Makes sense.

Opened https://github.com/varabyte/truthish/issues/13 and will add a commit for it it shortly.

I'll approve this PR now and leave it up to you if you want to wait for a new version of truthish.

common/kobwebx-frontmatter/src/commonTest/kotlin/com/varabyte/kobwebx/markdown/FrontMatterElementTest.kt
```diff
@@ -56,10 +58,10 @@ class FrontMatterElementTest {
         assertThat(fmElement["data.assets.font.Roboto"]).isNull()
 
         // Ensure getValue throws or not throws (as expected)
-        fmElement.getValue("title")
-        fmElement.getValue("data.assets.images")
-        fmElement.getValue("data.assets.images.0")
-        fmElement.getValue("data.assets.font")
+        val _ = fmElement.getValue("title")
```
bitspittle, line 59: Curious on your thoughts. I feel using `val _ = ...` everywhere is kind of ugly _in test code_ (I think it's really good to have it in non-test code for sure though).

I guess I see three paths:

1) Find a way to disable return value checking for code under test (may require reaching out to the kotlin team with feedback, as I'm not sure if it's currently possible)
2) Just use the IgnoreReturnValue annotation more aggressively
3) Keep the `val _` code even in tests, because it does technically communicate something and maybe it's not as ugly as I think.

bitspittle (approved):

common/kobwebx-frontmatter/src/commonTest/kotlin/com/varabyte/kobwebx/markdown/FrontMatterElementTest.kt
DennisTsar: I actually don't think it's ugly. Discarding values is rare even in test code, so it makes sense to me to mark it explicitly. I would definitely want a warning if I called `assertThat(...)` and did nothing with it (though this requires truthish to opt in).
Maybe the "proper" fix here is to have an `assertSucceeds` function.

backend/io/src/test/kotlin/com/varabyte/kobweb/io/ByteSourceTest.kt
bitspittle: Truthish 1.0.4 is out! So let's fold that change into this PR if it's not too much trouble?

DennisTsar: Done!

common/kobwebx-frontmatter/src/commonTest/kotlin/com/varabyte/kobwebx/markdown/FrontMatterElementTest.kt
bitspittle: OK, I'm convinced. Let's keep it this way. As long as it is relatively rare, then seeing `val _ = ...` adds signal and isn't just noise.

tools/ksp/site-processors/src/main/kotlin/com/varabyte/kobweb/ksp/frontend/FrontendProcessor.kt
```diff
@@ -773,7 +771,7 @@ class FrontendProcessor(
         }
 
         // Check for cycles starting from each layout
-        layoutsFor.keys.forEach { layout -> detectLayoutCycle(layout) }
+        layoutsFor.keys.forEach { layout -> val _ = detectLayoutCycle(layout) }
```
bitspittle: Make detectLayoutCycle ignore return value?

frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/ui/Modifier.kt
```diff
@@ -72,6 +72,7 @@ fun Modifier.thenUnless(condition: Boolean, other: Modifier): Modifier {
  */
 @OptIn(ExperimentalContracts::class, ExperimentalExtendedContracts::class)
 inline fun Modifier.thenIf(condition: Boolean, lazyProduce: () -> Modifier): Modifier {
+    @Suppress("RETURN_VALUE_NOT_USED") // `holdsIn` is incorrectly not marked as ignorable
```
bitspittle, line 75: Is JB aware of this?

frontend/silk-foundation/src/jsMain/kotlin/com/varabyte/kobweb/silk/style/CssStyle.kt
```diff
@@ -435,8 +435,8 @@ internal class SimpleCssStyle(
     extraModifier: @Composable () -> Modifier,
     val layer: String?
 ) : CssStyle<GeneralKind>(init, extraModifier) {
-    internal fun addStylesInto(styleSheet: StyleSheet): ClassSelectors {
-        return addStylesInto(selector, styleSheet, layer)
+    internal fun addStylesInto(styleSheet: StyleSheet) {
+        val _ = addStylesInto(selector, styleSheet, layer)
```
bitspittle, line 441: Maybe delegate addStylesInto should be ignore return value?

gradle/build-logic/publish/src/main/kotlin/com/varabyte/kobweb/gradle/publish/KobwebPublishPlugin.kt
```diff
@@ -85,6 +85,11 @@ class KobwebPublishPlugin : Plugin<Project> {
             apply("com.vanniktech.maven.publish")
         }
 
+        // Opt in published artifacts to unused return value checker
+        // See: https://github.com/Kotlin/KEEP/blob/main/proposals/KEEP-0412-unused-return-value-checker.md#opting-in-your-library
+        (project.kotlinExtension as HasConfigurableKotlinCompilerOptions<*>)
+            .compilerOptions.freeCompilerArgs.add("-Xreturn-value-checker=full")
```
bitspittle, line 91: How confident are we in doing a full mark of all Kobweb methods? Did you audit everything or should I also be looking through the whole API to verify this is a good change?

bitspittle (approved):

bitspittle:
BTW I'm thinking, and perhaps you came to the same conclusion, but we might want to hold off on this one until the next release?

frontend/kobweb-compose/src/jsMain/kotlin/com/varabyte/kobweb/compose/ui/Modifier.kt
DennisTsar: Yes but I didn't find a YT issue, only a Slack message: https://kotlinlang.slack.com/archives/C0B8Q383C/p1765079979003279?thread_ts=1765079979.003279&cid=C0B8Q383C

frontend/silk-foundation/src/jsMain/kotlin/com/varabyte/kobweb/silk/style/CssStyle.kt
DennisTsar: The returned `ClassSelectors` is important in the general case as without it the correct set of class names can't be applied to the target element. It's only in the specific case of `SimpleCssStyle` that the return value doesn't matter since it comes from a plain `registerStyle("...") {}` call that doesn't have any class names - I can add a comment explaining that.

bitspittle: Maybe yeah. "Don't need to pass on the return value because..." may become a common pattern, we'll see! :)

gradle/build-logic/publish/src/main/kotlin/com/varabyte/kobweb/gradle/publish/KobwebPublishPlugin.kt
DennisTsar: I did look through everything, both during the 2.3 beta and again after the release. 
It's still very possible I missed some functions, but I don't think it's a big deal if so - at worst it's a suppressible warning for users that opt in to an experimental feature.
For that reason I think it's safe to include this in the upcoming release, but your call about that and whether you want to do your own check through the whole API.

bitspittle: OK let's go for it :)

tools/ksp/site-processors/src/main/kotlin/com/varabyte/kobweb/ksp/frontend/FrontendProcessor.kt
DennisTsar: If you think that's nicer than ignoring it in one spot, sure. Since it's a local function I just went for the simpler option.

bitspittle (approved):
OK why not, let's go for it :) Let's go ahead and merge.

I actually agreed to that in a comment I left on Dec 21st so who am I not to be true to myself?

PR #766, opened 2026-01-10
Improve text escaping in markdown handlers

DennisTsar:


frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/dom/GenericTag.kt
```diff
@@ -132,3 +132,15 @@ fun <E : Element> GenericTag(
         content
     )
 }
+
+// Like the previous GenericTag, but not generic so the `attrs` lambda can be used without specifying a type
```
bitspittle: How about:

```
/**
 * A convenience version of [GenericTag] with a default template type specified.
 */
```

but would it be so bad to generate `GenericTag<Element>` in the code from markdown?

tools/gradle-plugins/extensions/markdown/src/main/kotlin/com/varabyte/kobwebx/gradle/markdown/handlers/MarkdownHandlers.kt
```diff
@@ -342,6 +344,20 @@ abstract class MarkdownHandlers @Inject constructor(project: Project) {
...
 
+        // Convert a set of HTML attributes to an `AttrBuilderContext` lambda block.
+        fun Attributes.toAttrsBlock(): String {
+            val styleMap = this.takeUnless { it.isEmpty } ?: return "{}"
+            return buildString {
+                append("{")
+                append(styleMap.joinToString(";") { (key, value) ->
+                    // Attribute values may contain meaningful new lines, e.g. `title="line1\nline2"`
+                    // These values are generally not long, so for simplicity we don't use multiline strings
+                    "attr(\"$key\", \"${value.escapeSingleQuotedText().replace("\n", "\\n")}\")"
```
bitspittle: Optional: Should we add "escapeNewline"? For consistency with the other escape methods in here

bitspittle (approved):

frontend/compose-html-ext/src/jsMain/kotlin/com/varabyte/kobweb/compose/dom/GenericTag.kt
DennisTsar: `GenericTag<Element>` would make the code generation a bit uglier because it would involve either adding `<Element>` when there are no attributes and it's not needed, or having attributes-related logic in two places, both at the point of `<Element>` and at the actual `attrs` block.
But more importantly, I think having an overload that doesn't require specifying a type would be nice in general - that's how the `attrsStr` version already works and as I think the lambda block overload is a strictly better version of that.

DennisTsar: I ended up changing my mind, so now it generates `GenericTag<org.w3c.dom.Element>` unconditionally for simplicity.

bitspittle (approved):

PR #771, opened 2026-02-07
Fix `Modifier.displayBetween` and add tests

DennisTsar:


bitspittle (approved):
Thanks! (Although I'm missing exactly what was broken before in the left / right comparison, maybe add a bit more info to the commit message?)

DennisTsar:
The first commit has the before/after of the fix: https://github.com/varabyte/kobweb/pull/771/changes/af2862a59f3d4851db037b030baccf41b3f629cb

PR #772, opened 2026-02-11
Add a warning if an optional segment isn't necessary due to a static route

bitspittle:


tools/processor-common/src/main/kotlin/com/varabyte/kobweb/project/frontend/FrontendData.kt
```diff
@@ -73,6 +77,33 @@ private fun Iterable<PageEntry>.assertValidPages(throwError: (String) -> Unit) {
...
         }
     }
+
+    // Collect all entries that end with an optional segment, e.g. "/a/b/c/{d?}", as we'll check if their parent dir is
+    // already registered separately, e.g. "/a/b/c/". If so, that means the former route can probably drop the ?, e.g.
+    // "/a/b/c/{d}", as the empty case will always be handled by the static route.
+    this.filter { entry ->
+        // Only the last segment in the route can be optional, so we only have to check that one
+        DynamicRouteSegment.tryCreate(entry.route.substringAfterLast('/'))?.isOptional ?: false
+    }.let { catchAllEntries ->
```
DennisTsar: The logic looks good, but I think it would be more readable if the `let` was dropped either by chaining the `forEach` or assigning the list to a variable. The `if` statement is also redundant.

bitspittle: Ah yeah thanks. I had an intermediate version I was playing around with where I was using catchAllEntries, but in the end it doesn't look like it's necessary.

tools/processor-common/src/main/kotlin/com/varabyte/kobweb/project/frontend/FrontendData.kt
```diff
@@ -73,6 +77,28 @@ private fun Iterable<PageEntry>.assertValidPages(throwError: (String) -> Unit) {
...
+    // Collect all entries that end with an optional segment, e.g. "/a/b/c/{d?}", as we'll check if their parent dir is
+    // already registered separately, e.g. "/a/b/c/". If so, that means the former route can probably drop the ?, e.g.
+    // "/a/b/c/{d}", as the empty case will always be handled by the static route.
+    this.filter { entry ->
+        // Only the last segment in the route can be optional, so we only have to check that one
+        DynamicRouteSegment.tryCreate(entry.route.substringAfterLast('/'))?.isOptional ?: false
+    }.forEach { entry ->
+        val parentDir = entry.route.substringBeforeLast('/').anonymizeDynamicSegments() + "/"
+        val finalSegment = entry.route.substringAfterLast('/')
+        val matchingStateEntry = entriesByRoute[parentDir]?.first()
```
DennisTsar: `matchingStateEntry` is a strange name, `matchingRouteEntry` maybe?

DennisTsar (approved):

tools/processor-common/src/main/kotlin/com/varabyte/kobweb/project/frontend/FrontendData.kt
bitspittle: Whoops! It was supposed to be "matchingPageEntry" -- using "state" was just a brain fart. Thanks!

PR #775, opened 2026-02-19
Add Lucide Icons

Ayfri:
This pull request introduces a new package for integrating Lucide icons into Kobweb projects, along with documentation and an automated icon generation system.

**Lucide Icon Integration and Automation**

* Added a `build.gradle.kts` script in `frontend/silk-icons-lucide` that parses a list of Lucide icons and generates corresponding Kotlin composable functions, including handling for deprecated icon aliases. This enables automatic code generation for icon wrappers and ensures the icon set stays up to date.
* Provided detailed documentation in `frontend/silk-icons-lucide/README.md` explaining how to integrate Lucide icons, load the required JS library, use the generated composables, and regenerate the icon source files after updating the icon list.


bitspittle:
Let me start with a high level comment first. **This isn't an action item for change as much as an opportunity for a quick discussion.**

I decided to look how the other PR approached their solution, since I don't recall them having issues with when to call the `createIcon` method, and here is the first few hundred lines of generated code that looks like this:

```kotlin
// !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
// THIS FILE IS AUTOGENERATED.
//
// Do not edit this file by hand. Instead, update `lucide-icons.json` in the module root and run the Gradle
// task "generateIcons"
// !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

package com.varabyte.kobweb.silk.components.icons.lucide

import androidx.compose.runtime.Composable
import com.varabyte.kobweb.compose.ui.Modifier
import com.varabyte.kobweb.compose.ui.toAttrs
import com.varabyte.kobweb.compose.dom.svg.Circle
import com.varabyte.kobweb.compose.dom.svg.Line
import com.varabyte.kobweb.compose.dom.svg.Path
import com.varabyte.kobweb.compose.dom.svg.Polygon
import com.varabyte.kobweb.compose.dom.svg.Polyline
import com.varabyte.kobweb.compose.dom.svg.Rect
import com.varabyte.kobweb.compose.dom.svg.ViewBox
import com.varabyte.kobweb.silk.components.icons.createIcon
import com.varabyte.kobweb.silk.components.icons.IconRenderStyle
import org.jetbrains.compose.web.css.*

@Composable
private fun renderIcon(
    elements: List<Pair<String, Map<String, String>>>,
    modifier: Modifier = Modifier,
    size: CSSLengthValue = 1.em,
    strokeWidth: Number = 2,
    color: CSSColorValue? = null
) {
    createIcon(
        viewBox = ViewBox.sized(24),
        width = size,
        renderStyle = IconRenderStyle.Stroke(strokeWidth),
        attrs = modifier.toAttrs {
            if (color != null) {
                attr("stroke", color.toString())
            }
            attr("fill", "none")
            attr("stroke-linecap", "round")
            attr("stroke-linejoin", "round")
        }
    ) {
        elements.forEach { (elementType, attributes) ->
            when (elementType) {
                "path" -> {
                    Path {
                        attributes["d"]?.let { d(it) }
                    }
                }
                "circle" -> {
                    Circle {
                        attributes["cx"]?.let { cx(it.toDoubleOrNull() ?: 0.0) }
                        attributes["cy"]?.let { cy(it.toDoubleOrNull() ?: 0.0) }
                        attributes["r"]?.let { r(it.toDoubleOrNull() ?: 0.0) }
                    }
                }
                "rect" -> {
                    Rect {
                        attributes["x"]?.let { x(it.toDoubleOrNull() ?: 0.0) }
                        attributes["y"]?.let { y(it.toDoubleOrNull() ?: 0.0) }
                        attributes["width"]?.let { width(it.toDoubleOrNull() ?: 0.0) }
                        attributes["height"]?.let { height(it.toDoubleOrNull() ?: 0.0) }
                        attributes["rx"]?.let { rx(it.toDoubleOrNull() ?: 0.0) }
                        attributes["ry"]?.let { ry(it.toDoubleOrNull() ?: 0.0) }
                    }
                }
                "line" -> {
                    Line {
                        attributes["x1"]?.let { x1(it.toDoubleOrNull() ?: 0.0) }
                        attributes["y1"]?.let { y1(it.toDoubleOrNull() ?: 0.0) }
                        attributes["x2"]?.let { x2(it.toDoubleOrNull() ?: 0.0) }
                        attributes["y2"]?.let { y2(it.toDoubleOrNull() ?: 0.0) }
                    }
                }
                "polyline" -> {
                    Polyline {
                        attributes["points"]?.let { pointsStr ->
                            // Parse "x1,y1 x2,y2 ..." format
                            val pairs = pointsStr.split(" ").mapNotNull { point ->
                                val coords = point.split(",")
                                if (coords.size == 2) {
                                    val x = coords[0].toDoubleOrNull()
                                    val y = coords[1].toDoubleOrNull()
                                    if (x != null && y != null) x to y else null
                                } else null
                            }
                            if (pairs.isNotEmpty()) {
                                points(*pairs.toTypedArray())
                            }
                        }
                    }
                }
                "polygon" -> {
                    Polygon {
                        attributes["points"]?.let { pointsStr ->
                            // Parse "x1,y1 x2,y2 ..." format
                            val pairs = pointsStr.split(" ").mapNotNull { point ->
                                val coords = point.split(",")
                                if (coords.size == 2) {
                                    val x = coords[0].toDoubleOrNull()
                                    val y = coords[1].toDoubleOrNull()
                                    if (x != null && y != null) x to y else null
                                } else null
                            }
                            if (pairs.isNotEmpty()) {
                                points(*pairs.toTypedArray())
                            }
                        }
                    }
                }
            }
        }
    }
}

@Composable
fun LiAArrowDown(
    modifier: Modifier = Modifier,
    size: CSSLengthValue = 1.em,
    strokeWidth: Number = 2,
    color: CSSColorValue? = null
) {
    renderIcon(
        listOf("path" to mapOf("d" to "m14 12 4 4 4-4"), "path" to mapOf("d" to "M18 16V7"), "path" to mapOf("d" to "m2 16 4.039-9.69a.5.5 0 0 1 .923 0L11 16"), "path" to mapOf("d" to "M3.304 13h6.392")),
        modifier, size, strokeWidth, color
    )
}

// etc.
```

In other words, they bake the SVG logic inside our code.

Now that you've done it a different way and are closer to the code than I am, do you have any thoughts about this approach?

Ayfri:
We could use this solution too, it's more similar to how Lucide handles the different frameworks yes, I choose to took the simplest way first, if you agree, I can refactor to use this solution!

bitspittle:
Honestly, I was hoping you would push back and say your way is much better! Or list the pros and cons of both approaches?

bitspittle:
Oh also, one thing to know about this solution, is the way kotlin/js works that LoP found out was if you include one function inside a file, you kind of pull them all in. The dead code elimination step can't remove all the other functions, because they're treated as methods inside an anonymous class.

That's why, if you look at our source code, under the SVG package, every single function is in its own file.

If we went with a similar approach here, we'd have to generate one file per function. Otherwise, people including Lucide icons would get slammed with all the data.

Ayfri:
I think I'm going to refactor the code to use a similar code than the previous PR to fix those issues. Yeah I know that it would import all icons but for Font Awesome it's importing a font containing all of the icons too, and that's okay for many people.

frontend/silk-icons-lucide/build.gradle.kts
```diff
@@ -0,0 +1,421 @@
...
+            "rect" to listOf("x", "y", "width", "height", "rx", "ry"),
+        )
+
+        fun generateElementCode(tag: String, attributes: Map<String, String>): String {
+            val tagCapitalized = tag.replaceFirstChar { it.uppercase() }
+
+            // Path has unique code generation (string attribute with escaping)
+            if (tag == "path") {
+                val d = attributes["d"] ?: ""
+                return "        Path {\n            d(\"${escapeKotlin(d)}\")\n        }"
```
bitspittle: Can we make this line more readable? Maybe use a multiline string?
```kotlin
return 
"""
Path {
   ${escapeKotlin(d)}
}
"""
```

frontend/silk-icons-lucide/build.gradle.kts
```diff
@@ -0,0 +1,421 @@
...
+            }
+
+            // Simple elements: open tag, emit each known attribute if present, close tag
+            simpleElementAttrs[tag]?.let { attrNames ->
+                return buildString {
+                    append("        $tagCapitalized {\n")
```
bitspittle: It seems a little hard to follow to be hiding indentation like this across string constants. 

One approach here is to put things behind a variable? `private val indent = "   "` somewhere?

Another is to create data structures of values, collect them in a first pass, and then convert those data values into strings on a second pass, at which point you start inserting things into indented raw string templates.

Like, imagine a split like this: 

```kotlin
class AttrInfo(val tag: String, val name: String)

// Parse attributes map and create a list of attr infos

buildString {
   append("${indent]Path {")
   for (attr in attrInfo) {
      append("${indent}${indent}${attrInfo.tag.capitalize()}")
   }
}
```

In other words, replace all these smaller buildstring calls with a single buildstring call at the end, and capture things into rich typed classes. I can't say I'm 100% sure but I bet if you did it this would read a lot better / be easier to modify in the future if requirements change.

(This is a very rough equivalent of the coding approach of splitting the model and the UI, if you think about it)

frontend/silk-icons-lucide/build.gradle.kts
```diff
@@ -0,0 +1,421 @@
...
+
+        // Maps SVG element tags to their known attribute names for code generation
+        val simpleElementAttrs = mapOf(
+            "circle" to listOf("cx", "cy", "r"),
+            "ellipse" to listOf("cx", "cy", "rx", "ry"),
+            "line" to listOf("x1", "y1", "x2", "y2"),
+            "rect" to listOf("x", "y", "width", "height", "rx", "ry"),
+        )
+
+        fun generateElementCode(tag: String, attributes: Map<String, String>): String {
```
bitspittle: It would be REALLY helpful to add a dev comment here (or maybe at the top of the class) that shows example input and example output. There's a lot of parsing code below and I could follow it a lot better if I understood the shape of the input coming in.

frontend/silk-icons-lucide/build.gradle.kts
```diff
@@ -0,0 +1,421 @@
...
+            ?: throw GradleException("Could not find 'icons' block in $generatedJsonFile")
+        val iconsBlock = iconsBlockMatch.groupValues[1]
+
+        val iconPattern = "\"([^\"]+)\"\\s*:\\s*\\[(.*)\\]".toRegex()
+        // icon name -> list of (tag, attrs) pairs
+        val activeIcons = mutableMapOf<String, List<Pair<String, Map<String, String>>>>()
+
+        for (line in iconsBlock.split("\n")) {
+            val trimmed = line.trim().removeSuffix(",")
+            val match = iconPattern.matchEntire(trimmed) ?: continue
```
bitspittle: Does this failed match mean we're skipping over an icon silently?

frontend/silk-icons-lucide/build.gradle.kts
```diff
@@ -0,0 +1,421 @@
...
+        }
+
+        val jsonText = inputJsonFile.get().asFile.readText()
+
+        val iconsBlockMatch = "\"icons\"\\s*:\\s*\\{(.*?)\\}\\s*,\\s*\"deprecated\"".toRegex(RegexOption.DOT_MATCHES_ALL)
+            .find(jsonText)
+            ?: throw GradleException("Could not find 'icons' block in $generatedJsonFile")
+        val iconsBlock = iconsBlockMatch.groupValues[1]
+
+        val iconPattern = "\"([^\"]+)\"\\s*:\\s*\\[(.*)\\]".toRegex()
```
bitspittle: Example input here would be helpful. Regexes are hard to read for squishy humans like me.

frontend/silk-icons-lucide/build.gradle.kts
```diff
@@ -0,0 +1,421 @@
...
+        val activeIcons = mutableMapOf<String, List<Pair<String, Map<String, String>>>>()
+
+        for (line in iconsBlock.split("\n")) {
+            val trimmed = line.trim().removeSuffix(",")
+            val match = iconPattern.matchEntire(trimmed) ?: continue
+            val iconName = match.groupValues[1]
+            val elementsStr = match.groupValues[2]
+
+            val elements = mutableListOf<Pair<String, Map<String, String>>>()
+            val elemPattern = "\\{\"tag\":\"([^\"]+)\"(?:,\"attrs\":\\{([^}]*)\\})?\\}".toRegex()
```
bitspittle: Ugh I'm realizing now there are a LOT of regexes with this approach. 

At the very least, collect these all to one place so we can reason about them together.

HOWEVER, regexes are almost always a code smell, and I have a growing feeling this is an indication that this approach is actually the wrong one, and your first approach may have been better. I would rather have a slightly inefficient payload that people won't even notice and simpler code than embedded SVGs that need to be regex parsed.  The other approach should never really go wrong, because all we're doing is downloading their code; but if this approach goes wrong in the future, I really don't want to be on the hook for fixing or adding new regexes...

frontend/silk-icons-lucide/README.md
bitspittle, line 1: So it feels a bit like AI generated this. If I'm wrong let me know. Be careful because AIs happily generate text that humans might not want to read or edit later. At least here it seems you prompted it to keep it short. 

The purpose of the README is really just for devs to understand the minimum of what's going on here and learn how to update things if users are asking for a new version in the future. You do have that in the latter half, everything after the `---`, so that's good. But you can definitely drop the first part about the icon parameters, which can easily go stale and it's not necessary here.

Here's how I might fix it:

```markdown
Support for integration of [Lucide icons](https://lucide.dev/) in your Kobweb project.

This directory contains a file called `lucide-icons.json`, which is parsed and used to generate code
used in this project.

To update it:

\`\`\`bash
./gradlew :frontend:silk-icons-lucide:fetchLucideIcons
\`\`\`

Once updated, run the following to regenerate the Kotlin source:

\`\`\`bash
./gradlew :frontend:silk-icons-lucide:generateIcons
\`\`\`

> [!NOTE]
> Each icon is generated as a separate Kotlin file to enable dead code elimination in Kotlin/JS -
> only the icons you actually use will be included in the final bundle.
```

frontend/silk-icons-lucide/README.md
```diff
@@ -0,0 +1,41 @@
...
+```kt
+LucideHome(size = 32.px, color = Colors.Red, strokeWidth = 1.5)
+```
+
+---
+
+Note that this directory contains a file called `lucide-icons.json`, which is parsed and used to generate code
+used in this project.
+
+Each icon is generated as a separate Kotlin file to enable dead code elimination in Kotlin/JS -
```
bitspittle: I'm glad this comment is here but I would also duplicate it in the build.gradle.kts task -- because that's probably where future me will go digging around first when trying to figure out the code.

tools/gradle-plugins/application/src/main/kotlin/com/varabyte/kobweb/gradle/application/tasks/KobwebGenerateSiteIndexTask.kt
```diff
@@ -18,6 +18,7 @@ import kotlinx.html.dom.document
 import kotlinx.html.dom.serialize
 import kotlinx.html.head
 import kotlinx.html.link
+import kotlinx.html.script
```
bitspittle: Nit: Remove?

bitspittle (commented):
Hey, so I finally started going through this, and I feel bad, because my gut it telling me this SVG approach is a mistake -- the sort of solution that an AI would happily push one into (because AIs don't fear regexes) but not what a human might have done on their own :).

My comments below are really at this point just concerns about readability and maintainability, nothing with functionality (which seems fine).

I'm not sure if it will be easier to continue moving forward through GitHub review or maybe you can reach out to me on Discord, so we can sort out any final points of friction.

I really appreciate you submitting this PR and I don't want the experience to be frustrating!

tools/gradle-plugins/application/src/main/kotlin/com/varabyte/kobweb/gradle/application/tasks/KobwebGenerateSiteIndexTask.kt
Ayfri: Yup!

frontend/silk-icons-lucide/README.md
Ayfri: Will do!

frontend/silk-icons-lucide/build.gradle.kts
Ayfri: Done!

frontend/silk-icons-lucide/build.gradle.kts
Ayfri: I've done something with a sealed class approach for different type of elements.

frontend/silk-icons-lucide/README.md
Ayfri: I see, no problem, I've simplified the README!

frontend/silk-icons-lucide/build.gradle.kts
Ayfri: I've used regexes since a few years and they aren't that hard in my opinion, also those are used in our own generated data, they'll never break by themselves. I've regrouped all of the regex patterns in one place and added comments to each of them.

frontend/silk-icons-lucide/build.gradle.kts
Ayfri: I've added a `logger.warn()`.

bitspittle (changes requested):
Hey Ayfri, and again, sorry for the huge delay on this. I really appreciate your patience.

The cleanup looks good. I find the Gradle code much easier to read right now.

I'm willing to roll with the regex approach for now and we can see how it works out. After all, worst case we can always switch over to another approach later; for the end user, they shouldn't notice. And this is a leaf module, not some core module that other code depends on.

**The only critical issue at the moment is that Widgets.kt doesn't compile right now (you just need to do an "optimize imports" in there). Fix that and I think we're good to go.**

**And optional, but it might also be nice if there was an SVG parsing library out there we could use and sidestep needing to rely on regexes entirely, just because parsing SVG code doesn't seem like something we should be getting into if we can avoid it.**

---

All that said, since I'm an old dinosaur and can't help myself, let me share some thoughts and advice that you are welcome to ignore :)

I still think you may be underestimating the pain of Regexes, and I only say this because there was a time I used to as well, until it bit me hard. 

To start.... if you take a step back, the average developer can't look at `\\{\"tag\":\"([^\"]+)\"(?:,\"attrs\":\\{([^}]*)\\})?\\}` and confirm if it's correct, or incomplete, or imbalanced, and also which groups match to which captured values. Maybe others are smarter than me, but my eyes cross pretty quickly looking at stuff like that. (The comments help a lot but could possibly go stale if people modify things in the future.)

The case I fear is that at some point in a year or two, someone asks us to update to Lucine Icons 2.0.0, and let's say the process just doesn't work, because the Lucide folks end up using some new SVG functionality. And now the next developer (probably me!) needs to unravel what the regexes you used were and how they should be modified or if new ones need to be added. Maybe they'll find this PR and need to do some archeology over our decisions, which will take time.

So, to be honest, I do have the vague sense we are putting a time bomb in our code. But, it might never ever go off! Or it might go off but just be a whimper, no big deal. So it's hard to push back too strongly.

But if something does goes wrong, this can be some of the worst kind of work to get stuck with, at least from my experience :) Because busted regexes are not really fun code to look into AND at the same time the issue becomes a huge priority that needs to be solved ASAP or else we disable lucide icons until we can resolve it. (In contrast, if we went with the original approach that puts a css link into the user's site, that should be much more future proof)

Ayfri:
I fixed the error!
Regarding the SVG parsing library, I agree with you but I have a different opinion.
SVG isn't that complex, is not changing very quickly and Lucide isn't changing quickly either, they modify some icons from time to time, but if all of the icons work today and regarding their code/design, I don't think they'll ever use a new syntax we don't support, and if they do, it might be simple to update.
The design is minimalist, simple and consistant, for me it's okay to follow that path.

If you really want an SVG parser doing the job more securely I totally understand and I can switch if necessary, but for now everything works so I don't see the point to change right now!

bitspittle:
Thank you Ayfri!!

PR #778, opened 2026-04-13
Add silk-icons-ms module for Material Symbols (#471)

### LINKED ISSUE #471: Support for Material Symbols
Kobweb supports MDI (material design icons), but Material Symbols is the new hotness, it seems.

See also: https://developers.google.com/fonts/docs/material_symbols

Instead of adding this

`<link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Material+Icons&family=Material+Icons+Outlined&family=Material+Icons+Two+Tone&family=Material+Icons+Round&family=Material+Icons+Sharp" />`

we need to add this:

`<link href="https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined" rel="stylesheet" />`

The best way forward here is probably forking the `silk-icons-mdi` project into a `silk-icons-ms` project (and then de-emphasizing mdi icons from the README?)

But this will require a bit of research first, to 1) see if this is really that different from MDI to support the effort and 2) see how much time it will take to fetch / create the symbols version of https://github.com/varabyte/kobweb/blob/main/frontend/silk-icons-mdi/md-icon-list.txt (which was done with a lot of inspecting webpages and manual reformatting of the output)

digitalby:
Fixes #471.

## Summary

Adds a new `silk-icons-ms` module that wraps Google's [Material Symbols](https://developers.google.com/fonts/docs/material_symbols) icon system, forked from `silk-icons-mdi` as suggested in the issue.

- New module at `frontend/silk-icons-ms`, published as `com.varabyte.kobwebx:silk-icons-ms`.
- Generates a `MsIcon(name, modifier, style)` composable plus ~4200 per-icon wrappers like `MsHome(style = MsIconStyle.ROUNDED)`. Three styles: `OUTLINED` (default), `ROUNDED`, `SHARP`.
- Uses an enum name `MsIconStyle` (rather than mirroring MDI's `IconStyle`) so both modules can be imported in the same file without needing an alias.
- Renders as `<span class="material-symbols-{outlined,rounded,sharp}">name</span>`, matching Google's documented markup.
- `KobwebGenerateSiteIndexTask` now injects the Material Symbols stylesheet link when the module is present, mirroring the existing MDI / FA branches.
- Playground updated with a "Web - Google Material Symbols" demo section in `Widgets.kt`.

## Icon list generation

`ms-icon-list.txt` is produced by a new `fetchMsIcons` Gradle task that downloads the three variable-font codepoint files from a pinned commit of [`google/material-design-icons`](https://github.com/google/material-design-icons) and formats them into the same style-grouped format used by MDI. Following the pattern of `silk-icons-lucide`'s `fetchLucideIcons`. Current pin: `229d4d6de51043272fdbb0137652f18f5ff4034b` (2026-04-10).

Refreshing the list in the future:

```bash
./gradlew :frontend:silk-icons-ms:fetchMsIcons
./gradlew :frontend:silk-icons-ms:generateIcons
```

## What's NOT in this PR (intentional scope cuts)

- **Variable font axes** (`FILL`, `wght`, `GRAD`, `opsz`). v1 renders at each font's default. Happy to file a follow-up for this if you want.
- **Deprecating or de-emphasizing `silk-icons-mdi`**. Kept as-is.
- **Root `README.md` changes**. Only the module-local README is added.
- **Aliasing `MsIconStyle` to the existing `IconStyle` type**. Kept distinct.

## Pre-existing bug I noticed while testing

When both `silk-icons-mdi` and `silk-icons-ms` are used together with `interceptUrls { enableSelfHosting() }`, the two Google Fonts `css2?family=...` URLs get rewritten to the same self-host path (`/_kobweb/self-host/fonts.googleapis.com/css2.css`), and one of the two CSS responses overwrites the other on disk. The head injection itself works correctly (verified via `URL intercepted:` log lines showing both URLs rewritten). Happy to file this as a separate issue if you confirm it reproduces on your end, but I didn't want to expand this PR's scope.

## Test plan

- [x] `./gradlew :frontend:silk-icons-ms:fetchMsIcons` downloads codepoints and writes `ms-icon-list.txt` (4220 icons per style).
- [x] `./gradlew :frontend:silk-icons-ms:generateIcons` produces `MsIcons.kt` under `build/generated/icons/src/jsMain/kotlin/...`.
- [x] `./gradlew :frontend:silk-icons-ms:build` compiles the generated Kotlin.
- [x] `./gradlew :tools:gradle-plugins:application:build` passes with the new `hasMsDependency` branch.
- [x] From `playground/`, `./gradlew :site:compileKotlinJs` succeeds with the new `Ms*` calls in `Widgets.kt`.
- [x] `./gradlew :site:kobwebGenSiteIndex --rerun-tasks` logs both `URL intercepted: "https://fonts.googleapis.com/css2?family=Material+Icons..."` and `URL intercepted: "https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined..."`, confirming head injection runs when the dep is present.
- [x] (reviewer) `cd playground/site && kobweb run`, open the Widgets page in a browser, and visually confirm `MsSearch()`, `MsHome(style = MsIconStyle.ROUNDED)`, `MsMenu(style = MsIconStyle.SHARP)`, and `MsClose()` all render as expected glyphs next to the existing MDI demo.

## Commits

- `feat: scaffold silk-icons-ms module to host Material Symbols icons`
- `feat: populate silk-icons-ms from Material Symbols codepoints`
- `feat: inject Material Symbols stylesheet when silk-icons-ms is present`
- `chore: smoke-test silk-icons-ms from playground site`
- `docs: add silk-icons-ms README covering install, styles, and regen`


frontend/silk-icons-ms/build.gradle.kts
```diff
@@ -0,0 +1,223 @@
...
+                    styles.size == 1 -> {
+                        "@Composable fun $methodName(modifier: Modifier = Modifier) = MsIcon(\"$rawName\", modifier, ${styles.first().name})"
+                    }
+
+                    styles.isEmpty() -> {
+                        // This shouldn't be possible, but just in case...
+                        throw GradleException("Unexpected icon entry with no styles: $entry")
+                    }
+
+                    styles.size == IconStyle.values().size -> {
```
bitspittle: Super nit: `IconStyle.values().size` -> `IconStyle.entries.size`

bitspittle (commented):
At first glance, the code seems great!

Main issue is when I check the widgets page I get this:

Image

Do you know why it might not be working for me? Even when I refresh the page I still see it.

---

Also! Do you think we should be updating the MDI fonts to migrate over to the new CDN location? That would help with your concern about duplicate URLs in enable self hosting, right? https://pictogrammers.com/docs/library/mdi/getting-started/webfont/

bitspittle:
I just reran the site after commenting out `enableSelfHosting` and sure enough that fixes the font issue, so I suspect this is the duplicate URL issue you were talking about, yeah?

digitalby:
> I just reran the site after commenting out `enableSelfHosting` and sure enough that fixes the font issue, so I suspect this is the duplicate URL issue you were talking about, yeah?

Yes, that's the same self-host duplicate-URL bug I flagged in the PR description. Both `silk-icons-mdi` and `silk-icons-ms` currently inject stylesheets under `fonts.googleapis.com/css2?...`, so the interceptor rewrites both to `/_kobweb/self-host/fonts.googleapis.com/css2.css` and the second fetch overwrites the first on disk. The right place to fix it is the URL interceptor (disambiguate colliding host+path pairs, e.g. by query hash) rather than either icons module.

This would probably need a separate issue filed?

As per our discussion `silk-icons-mdi` (Pictogrammers' Material Design Icons) and `silk-icons-ms` (Google's Material Symbols) are two distinct icon ecosystems, not two CDNs for the same font. The icon names and coverage don't overlap, so migrating one toward the other isn't really on the table. If the MDI stylesheet URL needs revisiting that's worth its own issue, but it's orthogonal to the duplicate-URL collision above.

bitspittle:
Thanks again for your help and patience. This is now in!

Here's the parent commit which fixed the issue with `enableSelfHosting`, if you're curious: https://github.com/varabyte/kobweb/commit/9772873ef1d528d169f070d06e2fd53ee8102347

PR #782, opened 2026-05-30
feat: Add configurable rate limiting to Kobweb server

FegzDev:
Hi @bitspittle,

This PR introduces a configurable rate-limiting system to the Kobweb backend server to help protect endpoints from abuse.
 
By default, rate limiting is disabled, ensuring full backwards compatibility. When enabled, developers can define both a global configuration and specific route-level overrides directly via `conf.yaml`.

The currently supported algorithms are Token Bucket and Fixed Window with the possibility of adding more configurable algorithms.

## Example Configuration (`conf.yaml`)
```yaml
server:
  rateLimiting:
    enabled: true
    global:
      type: TOKEN_BUCKET
      key: FORWARD_IP
      capacity: 100
      tokens: 10
      duration: 1s
    routes:
      - path: /api/sensitive
        configuration:
          type: FIXED_WINDOW
          key: IP
          size: 5
          duration: 1s
          
I would appreciate your feedback on this.

### DIFF
--- a/backend/server/src/main/kotlin/com/varabyte/kobweb/server/Application.kt
+++ b/backend/server/src/main/kotlin/com/varabyte/kobweb/server/Application.kt
@@ -14,6 +14,7 @@ import com.varabyte.kobweb.server.api.SiteLayout
 import com.varabyte.kobweb.server.io.ServerStateFile
 import com.varabyte.kobweb.server.plugin.KobwebServerPlugin
 import com.varabyte.kobweb.server.plugins.configureHTTP
+import com.varabyte.kobweb.server.plugins.configureRateLimiting
 import com.varabyte.kobweb.server.plugins.configureRouting
 import com.varabyte.kobweb.server.plugins.configureSerialization
 import io.ktor.server.application.*
@@ -77,8 +78,8 @@ suspend fun main() {
         System.setProperty(
             "LOG_MAX_HISTORY",
             maxFileCount?.takeIf { it > 0 }?.toString()
-                // Note: The "9999" hack value below used to be Int.MAX_VALUE, but this caused logback to break,
-                // consuming 100% of the thread. ~30 years (9999 days) of logs should be enough for anyone.
+            // Note: The "9999" hack value below used to be Int.MAX_VALUE, but this caused logback to break,
+            // consuming 100% of the thread. ~30 years (9999 days) of logs should be enough for anyone.
                 ?: if (totalSizeCap?.takeIf { it.inWholeBytes > 0 } == null) "0" else "9999")
         System.setProperty("LOG_SIZE_CAP", totalSizeCap?.inWholeBytes?.toString() ?: "0")
     }
@@ -123,6 +124,7 @@ suspend fun main() {
     val engine = embeddedServer(Netty, port) {
         log.info("Initializing server engine for Kobweb project \"${conf.site.title}\"")
 
+        configureRateLimiting(conf.server.rateLimiting)
         configureRouting(appProperties, env, siteLayout, conf, globals, events)
         configureSerialization()
         configureHTTP(appProperties, env, conf)
@@ -161,6 +163,7 @@ suspend fun main() {
                         globals.status.clear() // If we happen to be showing a message when paused, cancel it!
                     }
                 }
+
                 is ServerRequest.ResumeClientEvents -> {
                     if (pausedCopy != null) {
                         globals.set(pausedCopy)
@@ -168,9 +171,11 @@ suspend fun main() {
                         pausedCopy = null
                     }
                 }
+
                 is ServerRequest.IncrementVersion -> {
                     targetGlobals().version++
                 }
+
                 is ServerRequest.SetStatus -> {
                     targetGlobals().status.apply {
                         text = request.message
--- a/backend/server/src/main/kotlin/com/varabyte/kobweb/server/plugins/RateLimiting.kt
+++ b/backend/server/src/main/kotlin/com/varabyte/kobweb/server/plugins/RateLimiting.kt
@@ -1,0 +1,112 @@
+package com.varabyte.kobweb.server.plugins
+
+import com.varabyte.kobweb.common.error.KobwebException
+import com.varabyte.kobweb.project.conf.Server
+import com.varabyte.kobweb.server.ratelimiting.FixedWindowAlgorithm
+import com.varabyte.kobweb.server.ratelimiting.RateLimitingAlgorithm
+import com.varabyte.kobweb.server.ratelimiting.TokenBucketAlgorithm
+import io.ktor.http.HttpHeaders
+import io.ktor.http.HttpStatusCode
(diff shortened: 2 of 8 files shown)


bitspittle:
Thanks for the PR and taking the time to help improve Kobweb! 

Heads up, I'm traveling through June 4 (and will need a few days after that to recover from jet lag). I'll try to get back to you shortly after that.

FegzDev:
Noted.

Looking forward to your feedback and the possible integration of this feature.

bitspittle:
Thanks for the ping! I should be able to look into it today.

As a heads up, and I'm saying this before I've looked at the code so these are all general thoughts, nothing specific about your actual PR, I need to learn generally more about rate limiting so I can understand what most people's config requirements are. We also have the Kobweb server plugin, where I'm pretty sure people can also configure rate limiting, so I need to think at a higher level what features go into the server and what don't that I leave to the plugin. I've avoided doing that so far just because there has always been other priorities!

And finally, a feature like this will also require me to add a section of documentation to this page: https://kobweb.varabyte.com/docs/concepts/server/fullstack

Now that I'm back from my trip, I'm trying to hunker down and get a release out soon as well. So depending on the complexity of what this feature implies, I may decide to get the release out first and then fold this info the next release.

Just wanted to give you a heads up that a feature like this is more than just saying yes or no to code looking good. Thank you for understanding in advance!

bitspittle:
Also, a question: can I ask you what drove you to add this feature? Were you working on a site of your own and found its lack problematic for your own project? 

Knowing a specific, concrete pain point can help me more easily prioritize getting this feature in as well as help me write the docs section.

FegzDev:
> Thanks for the ping! I should be able to look into it today.
> 
> As a heads up, and I'm saying this before I've looked at the code so these are all general thoughts, nothing specific about your actual PR, I need to learn generally more about rate limiting so I can understand what most people's config requirements are. We also have the Kobweb server plugin, where I'm pretty sure people can also configure rate limiting, so I need to think at a higher level what features go into the server and what don't that I leave to the plugin. I've avoided doing that so far just because there has always been other priorities!
> 
> And finally, a feature like this will also require me to add a section of documentation to this page: https://kobweb.varabyte.com/docs/concepts/server/fullstack
> 
> Now that I'm back from my trip, I'm trying to hunker down and get a release out soon as well. So depending on the complexity of what this feature implies, I may decide to get the release out first and then fold this info the next release.
> 
> Just wanted to give you a heads up that a feature like this is more than just saying yes or no to code looking good. Thank you for understanding in advance!

I understand.

FegzDev:
> Also, a question: can I ask you what drove you to add this feature? Were you working on a site of your own and found its lack problematic for your own project?
> 
> Knowing a specific, concrete pain point can help me more easily prioritize getting this feature in as well as help me write the docs section.

I didn't really run into a specific issue.

I noticed many backend frameworks have some rate limting support and thought it would be nice to implement something like this for kobweb. 

However, now that you mentioned it, the server plugin kind of already provides some support for this with [Ktor Rate Limiting](https://ktor.io/docs/server-rate-limit.html).

A key difference between this and that of Ktor is the support for different algorithms on a global or per endpoint basis.

It structure allows the framework to add support for multiple customizable rate limiting algorithms in the future and for end users to pick from a list of algorithms (Token Bucket, Leaky Bucket, Fixed Window or any rate limiting algorithm out there).

FegzDev:
> > Also, a question: can I ask you what drove you to add this feature? Were you working on a site of your own and found its lack problematic for your own project?
> > Knowing a specific, concrete pain point can help me more easily prioritize getting this feature in as well as help me write the docs section.
> 
> I didn't really run into a specific issue.
> 
> I noticed many backend frameworks have some rate limting support and thought it would be nice to implement something like this for kobweb.
> 
> However, now that you mentioned it, the server plugin kind of already provides some support for this with [Ktor Rate Limiting](https://ktor.io/docs/server-rate-limit.html).
> 
> A key difference between this and that of Ktor is the support for different algorithms on a global or per endpoint basis.
> 
> It structure allows the framework to add support for multiple customizable rate limiting algorithms in the future and for end users to pick from a list of algorithms (Token Bucket, Leaky Bucket, Fixed Window or any rate limiting algorithm out there).

Now that I think about it, this PR does not include an option to exclude certain sub paths of a given url.

This can be valid when a user wants the rate limiting applied to all /api/a/b paths but excluding /api/a/b/c/d.

Also, maybe pattern matching or regex can be applied to the urls for better configuration.

Do take a look at all these and let me know your feedback.

bitspittle:
So I looked at the code and I now see this would be signing on for more work than I first expected.

I originally assumed you added a dependency on some battle-tested code (like a ktor plugin or library) for this, but we can't afford being responsible for implementing rate limiting logic ourselves. (Similar, for example, if we decide to add encrypted endpoints, we don't want to implement SHA-256 ourselves.)

If someone in the future files a bug due to a subtle implementation issue, or someone requests a new algorithm, I don't think I'll have the time to prioritize it. I'm already spread so thin.

For the latter point, you are right -- making sure the algorithms apply to the right subset of endpoints will be necessary. There may be some prior art that can be copied from the redirect feature: https://kobweb.varabyte.com/docs/concepts/foundation/redirects but I'd need to think more about it. Probably something that should be hashed out in a document first before code is written.

But still... I don't want you to spend more time on a PR that I might still struggle to accept. For the next step, would you be willing to join our Discord server or Kotlin slack channel and ask if anyone needs this feature? I don't want to come across as harsh, but I'm always trying to minimize the amount of code we're responsible for. For the last four years, pretty much every feature has been driven by user need.

FlyingPig525:
In my (albeit small amount of) experience, it has always been a better choice in the end to just use external software (such as nginx) and configure rate limiting using that. They generally have more expansive configuration options and would be much faster than a system built into kobweb.

bitspittle:
> In my (albeit small amount of) experience, it has always been a better choice in the end to just use external software (such as nginx) and configure rate limiting using that. 

I would definitely push people at this point to use an external system if they had a choice for sure! And I think it's fair for you to recommend that.

At the same time, I'm also happy to think of a Kobweb server as a valid, useful solution for users making a first pass of their site. And if I learn that configuring rate limiting is important to enough such users, I'd be happy to support it (assuming the feature is solid / easy to maintain)

FlyingPig525:
> At the same time, I'm also happy to think of a Kobweb server as a valid, useful solution for users making a first pass of their site. And if I learn that configuring rate limiting is important to enough such users, I'd be happy to support it (assuming the feature is solid / easy to maintain)

As a general best practice, I would hope, if a Kobweb user gets to a point at which they need rate limiting, they would already be using such external systems. Rate limiting is a critical factor of a server, and therefore should probably be done by a highly tested piece of software. Failure of rate limiting (both blocking or allowing when it shouldn't be) can be catastrophic to the end-user's or the developer's experience. I've not looked into this PR, other than a quick skim, nor tested it, but I feel the responsibility it attempts to fulfill should be out of the scope of Kobweb.

Issue #795, opened 2026-09-06
Command + Click on mac opens the new page in the same tab


ShaHar91:
Using `Command` + click on a link (on a Mac) opens the link in the same tab. This should be opened in a new (unfocused) tab.

* OS: Mac
* Kobweb libs version: 0.23.2
* Kobweb CLI version: 0.9.18

PR #796, opened 2026-09-06
Add the support for the `Command` key to open in a new tab as well

ShaHar91:
I filed a bug with tag #795 which this PR fixes


bitspittle (approved):
Thank you!! And also for filing an accompanying bug as well.

You know what's crazy? I've been on a Mac for two years now and didn't realize this :)

Will submit and put out a new snapshot momentarily. 30 minutes after you see this message, you can use `0.25.2-SNAPSHOT` in your `libs.versions.toml` for kobweb if you'd like.

